PR #6, opened 2025-02-13
Quality Control pipeline for EEG data

estefanysuarez:
Added the quality module to evaluate the quality of EEG data. This modules includes the following submodules:
- metrics: includes different quality metrics
- montages + native_montages: set of functions to create MNE standard montages or custom ones
- raw: contains the BrainsetsRaw class, which is an extension of mne.io.RawArray class to be able to perform different preprocessing procedures on the data
- pipeline: contains the QCPipeline class. The main method takes in a BrainsetsRaw object, applies some preprocessing steps on the raw data and returns the quality scores

**NOTE:**  new functionality requires the following additional packages mne-icalabel==0.7.0 and meegkit


brainsets/quality/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .pipeline import DynamicQCPipeline
```
milosobral, line 1: I don't think relative imports are the best way to go, I would change these to something like:
`from brainsets.quality.pipeline import DynamicQCPipeline`

brainsets/quality/montages.py
```diff
@@ -0,0 +1,131 @@
...
+    and fiducial points. Channels in ch_names that don't exist in the 10-05 system
+    will be excluded from the resulting montage.
+    """
+    standard_1005 = mne.channels.make_standard_montage("standard_1005")
+    ch_pos_1005 = standard_1005.get_positions()["ch_pos"]
+    ch_pos_custom = {ch: ch_pos_1005[ch] for ch in ch_names if ch in ch_pos_1005}
+
+    custome_montage = mne.channels.make_dig_montage(
+        ch_pos=ch_pos_custom,
+        nasion=standard_1005.get_positions()["nasion"],
```
milosobral, line 88: This is very nitpicky but if you're gonna be calling a this function a bunch of times you might want to instead get it's output in a new variable and using that variable:
`standard_positions = mne.channels.make_standard_montage("standard_1005").get_positions()`

brainsets/quality/montages.py
```diff
@@ -0,0 +1,131 @@
...
+    Notes
+    -----
+    The function uses the standard_1005 montage as reference for channel positions
+    and fiducial points. Channels in ch_names that don't exist in the 10-05 system
+    will be excluded from the resulting montage.
+    """
+    standard_1005 = mne.channels.make_standard_montage("standard_1005")
+    ch_pos_1005 = standard_1005.get_positions()["ch_pos"]
+    ch_pos_custom = {ch: ch_pos_1005[ch] for ch in ch_names if ch in ch_pos_1005}
+
```
milosobral, line 85: I would probably add a warning or something here if some of the channels is not used, to make sure it's really evident that some of the channel names given are incompatible

brainsets/quality/pipeline.py
```diff
@@ -0,0 +1,683 @@
...
+# Make ABC abstract class
+from abc import ABC, abstractmethod
+
+logging.basicConfig(format="%(message)s\n", level=logging.INFO)  # Only show the message
+
+
+class QCPipeline(ABC):
+    @abstractmethod
+    def __init__(self) -> None:
+        super().__init__()
```
milosobral, line 35: I don't really think you need this if you are only defining the init method which all classes have in some way. You can directly make `BaseQCPipeline the abstract class

brainsets/quality/pipeline.py
```diff
@@ -0,0 +1,683 @@
...
+
+
+class BaseQCPipeline(QCPipeline):
+    """
+    Base class for Quality Control (QC) Pipelines.
+    This class provides a framework for setting up and running quality control
+    pipelines on EEG data. It includes methods for setting up channels, montages,
+    filters, and Independent Component Analysis (ICA) for artifact removal.
+    """
```
milosobral, line 44: I think it would be nice to have the docstring explain what each of the arguments are

brainsets/quality/pipeline.py
```diff
@@ -0,0 +1,683 @@
...
+        quality_check_ica: bool = True,
+        include_components: List[str] = ["brain", "other"],
+        plot_scores: bool = True,
+        random_state: Union[None, int, np.random.RandomState] = None,
+    ):
+
+        mne.set_log_level("ERROR")
+
+        # Pipeline setup
+        self.channels_rename = channels_rename
```
milosobral, line 69: I think you forgot this variable here, it's not used in `_to_standard_channel_names`

brainsets/quality/pipeline.py
```diff
@@ -0,0 +1,683 @@
...
+    pipelines on EEG data. It includes methods for setting up channels, montages,
+    filters, and Independent Component Analysis (ICA) for artifact removal.
+    """
+
+    def __init__(
+        self,
+        # data
+        channels: List[str],
+        channels_rename: Optional[Dict[str, str]] = None,
+        montage_name: str = "standard_1005",
```
milosobral, line 51: Having this be the default value is going to lead to some confusion as if the user provides a list of channels and leave this empty, the behavior of `make_montage` in montages.py will lead to the montage being the standard 1005

brainsets/quality/pipeline.py
```diff
@@ -0,0 +1,683 @@
...
+        # Rename channels
+        if self.channels_rename is not None:
+            raw.rename_channels(self.channels_rename)
+            logging.info(f"Renamed channels: {self.channels_rename}.")
+
+        # Set montage
+        raw.set_montage(self.montage)
+
+        # Detrend data
+        # raw.detrend()
```
milosobral, line 514: Maybe forgot to uncomment/remove here?

brainsets/quality/pipeline.py
```diff
@@ -0,0 +1,683 @@
...
+            scores.append(
+                self.run_slice(
+                    raw.copy().crop(tmin=start, tmax=end, include_tmax=include_end)
+                )
+            )
+            # except Exception as e:
+            #     logging.error(
+            #         f"Time slice {i + 1}: {start:.1f} to {end:.1f} seconds failed. \tError: {e}"
+            #     )
+            #     scores.append(None)
```
milosobral, line 540: Probably want to remove this as well if it's not going to be used

milosobral (changes requested):

PR #8, opened 2025-02-19
Fix raw/proc dir creation from `brainsets config`

vinamarora8:
When doing `brainsets config`, if I use `~/test/raw` and `~/test/processed` as the 2 directories, the script incorrectly creates the directories `./~/test/raw` and `./~/test/processed`. 
Note that this bug surfaces only if the directories are entered through the prompts.

This PR fixes the problem.


mazabou (approved):
LGTM! 

Note that we currently have no workflow for testing the brainsets cli. we should add more in the workflow.

PR #9, opened 2025-02-19
Per-dataset environments

vinamarora8:
**Problem**: Each dataset needs its own environment on top of the base `brainsets` environment. Our current solution is to install different environments for each dataset manually.

**This PR**: Introduces the idea of using `uv run --with <extra requirements> [command]` to create automatically create isolated environments for each dataset.  The extra packages needed for each dataset will be defined in the `Snakefile` and the creators of these `Snakefiles` are expected to add them.

Currently implemented only for the two datasets we have in the main branch. 

- Removed `dandi` as a requirement of `brainsets` and added `uv`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ dependencies = [
     "pydantic~=2.0",
     "pulp==2.7.0",
     "click~=8.1.3",
-    "dandi==0.61.2",
+    "uv",
 ]
 
 [project.optional-dependencies]
--- a/.gitignore
+++ b/.gitignore
@@ -161,3 +161,5 @@ data/raw
 # macOS system files
 .DS_Store
 **/.DS_Store
+
+uv.lock
--- a/brainsets_pipelines/pei_pandarinath_nlb_2021/Snakefile
+++ b/brainsets_pipelines/pei_pandarinath_nlb_2021/Snakefile
@@ -8,13 +8,15 @@ DANDI_ID = "000140"
 RAW_DIR = config["RAW_DIR"]
 PROCESSED_DIR = config["PROCESSED_DIR"]
 
+EXTRA_REQS = "dandi==0.61.2"
+
 checkpoint pei_pandarinath_nlb_2021_download_data:
     output:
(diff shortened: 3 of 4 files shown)


mazabou (commented):
Neat, is this environment created temporarily? and is it only valid within the context of that snakefile's script?

vinamarora8:
Yes, the environment is created only temporarily,  and valid only for the commands that are run with `uv run`.
Actually, give me another day to experiment with this and ensure it's all working as we would expect.

vinamarora8:
@mazabou Made the change we discussed offline. Ready for review.

pyproject.toml
mazabou, line 1: do we want to consider thinning out unnecessary packages that are specific to some datasets? pynwb, scipy, scikit-image etc..

happy to make this a different PR too

brainsets/cli.py
```diff
@@ -43,19 +43,38 @@ def prepare(dataset, cores):
...
+    if reqs_filepath.exists():
+        uv_prefix_command = [
+            "uv",
+            "run",
+            "--with-requirements",
+            str(reqs_filepath),
+            "--active",  # Prefer building temp environment on top of current venv
+        ]
+        command = uv_prefix_command + command
+        click.echo(f"Using extra requirements from {reqs_filepath}")
```
mazabou, line 72: the term extra requirements might be confusing? 
How about something that explains what is happening: `Building temporary virtual environment using requirements from ...`

pyproject.toml
vinamarora8: Agreed 100%, that would be super good.
Also we should reduce requirements of specific versions of packages.
But lets make this a separate PR.

vinamarora8: We can also add (for e.g.) specific versions of numpy in the dataset specific `requirements.txt` files. 
This would be good, since packages like numpy don't guarantee same RNG across versions.

vinamarora8:
@mazabou made the suggested edit

mazabou (approved):
LGTM!

PR #12, opened 2025-02-20
added new attributes to brainsetdescriptions

colehurwitz:
Added `publication_list` and `subject_list` to the BrainsetDescription. Will be helpful for automatically generating READMEs in future updates. Updated the perich dataset to show how this works. Also, added a test for descriptions.

### DIFF
--- a/brainsets/descriptions.py
+++ b/brainsets/descriptions.py
@@ -25,10 +25,15 @@ class BrainsetDescription(temporaldata.Data):
         Original data source (usually a URL, or a short description otherwise)
     description : str
         Text description of the brainset
+    publication_list : list[str]
+        List of references to publication associated with the dataset
+    subject_list : list[str]
+        List of subjects in the dataset
     brainsets_version : str, optional
         Version of brainsets package used, defaults to current version
     temporaldata_version : str, optional
         Version of temporaldata package used, defaults to current version
+
     """
 
     id: str
@@ -36,6 +41,8 @@ class BrainsetDescription(temporaldata.Data):
     derived_version: str
     source: str
     description: str
+    publication_list: list[str]
+    subject_list: list[str]
     brainsets_version: str = brainsets.__version__
     temporaldata_version: str = temporaldata.__version__
 
--- a/brainsets_pipelines/perich_miller_population_2018/prepare_data.py
+++ b/brainsets_pipelines/perich_miller_population_2018/prepare_data.py
@@ -159,6 +159,13 @@ def main():
         "contains spiking activity—manually spike sorted in three subjects, and "
         "threshold crossings in the fourth subject—obtained from up to 192 electrodes "
         "per session, cursor position and velocity, and other task related metadata.",
+        publication_list=[
+            "Perich MG, Gallego JA, Miller LE. A Neural Population Mechanism for Rapid Learning. Neuron. 2018 Nov 21;100(4):964-976.e7. doi: 10.1016/j.neuron.2018.09.030. Epub 2018 Oct 18. PMID: 30344047; PMCID: PMC6250582.",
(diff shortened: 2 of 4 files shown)


brainsets/descriptions.py
```diff
@@ -25,17 +25,24 @@ class BrainsetDescription(temporaldata.Data):
         Original data source (usually a URL, or a short description otherwise)
     description : str
         Text description of the brainset
+    publication_list : list[str]
+        List of references to publication associated with the dataset
+    subject_list : list[str]
+        List of subjects in the dataset
```
mazabou, line 31: In general, we don't want to define metadata redundantly. The list of subjects can be inferred from `SubjectDescription`.

brainsets/descriptions.py
```diff
@@ -25,17 +25,24 @@ class BrainsetDescription(temporaldata.Data):
...
     brainsets_version : str, optional
         Version of brainsets package used, defaults to current version
     temporaldata_version : str, optional
         Version of temporaldata package used, defaults to current version
+
```
mazabou, line 36: rogue line?

brainsets_pipelines/perich_miller_population_2018/prepare_data.py
```diff
@@ -159,6 +159,13 @@ def main():
...
         "per session, cursor position and velocity, and other task related metadata.",
+        publication_list=[
+            "Perich MG, Gallego JA, Miller LE. A Neural Population Mechanism for Rapid Learning. Neuron. 2018 Nov 21;100(4):964-976.e7. doi: 10.1016/j.neuron.2018.09.030. Epub 2018 Oct 18. PMID: 30344047; PMCID: PMC6250582.",
+            "Perich MG, Miller LE. Altered tuning in primary motor cortex does not account for behavioral adaptation during force field learning. Exp Brain Res. 2017 Sep;235(9):2689-2704. doi: 10.1007/s00221-017-4997-1. Epub 2017 Jun 6. PMID: 28589233; PMCID: PMC5709199.",
+            "Glaser JI, Perich MG, Ramkumar P, Miller LE, Kording KP. Population coding of conditional probability distributions in dorsal premotor cortex. Nat Commun. 2018 May 3;9(1):1788. doi: 10.1038/s41467-018-04062-6. PMID: 29725023; PMCID: PMC5934453.",
+            "Churchland MM, Cunningham JP, Kaufman MT, Foster JD, Nuyujukian P, Ryu SI, Shenoy KV. Neural population dynamics during reaching. Nature. 2012 Jul 5;487(7405):51-6. doi: 10.1038/nature11129. PMID: 22722855; PMCID: PMC3393826.",
+        ],
```
mazabou, line 167: we can limit this to doi. do you think we need to have a regex validator, to check that the doi is correct?

brainsets/descriptions.py
```diff
@@ -25,17 +25,24 @@ class BrainsetDescription(temporaldata.Data):
         Original data source (usually a URL, or a short description otherwise)
     description : str
         Text description of the brainset
+    publication_list : list[str]
```
mazabou, line 28: in addition to publication_list, do we want to add lab? or is that session specific? -- i guess it is for IBL data

mazabou (changes requested):

brainsets/descriptions.py
colehurwitz: Will remove. Thought black might catch these.

brainsets_pipelines/perich_miller_population_2018/prepare_data.py
colehurwitz: doi is a good idea. I can add a regex validator later when I build the validators.

brainsets/descriptions.py
colehurwitz: as you mentioned, lab is challenging for multi-lab datasets. Probably not worth it when we put the publication.

brainsets/descriptions.py
colehurwitz: Yep, I can remove this.

PR #14, opened 2025-02-21
Making brainsets installable from pip + In-house temporary `venv` management

vinamarora8:
### Overview
Currently, brainsets is runnable only in either editable-mode (i.e., `git clone`, then `pip install -e .`), or in non-editable-mode but when you're in the brainsets project directory (i.e. `git clone`, then `pip install .`).

These edits seem necessary for brainsets to be runnable from a direct pip install into a fresh venv:

1. (Minor) Add submodules of `brainsets` as packages to be installed in `pyproject.toml`
2. (Major) Custom temporary `venv`: Looks like `uv run` is not upto the mark for layering environments as well as we need it to (I experimented a lot, but might've missed something ofc). We didn't see any problems because we were installing brainsets in editable mode/running from within a brainsets clone. More about this near the end of this description.

### Custom temp `venv` solution
Follows these steps:
1. Find current brainsets installation using `pip freeze` 
    (could be `brainsets==0.1.2` or `brainsets @ git+https://...` or `-e file://local_brainsets_clone/`)
2. Create a temp venv in `<processed_dir>/<dataset>/tmp/venv`
3. Install previously found brainsets package
    (we need to install the exact version as current environment for things to remain seemless)
4. Install dataset specific requirements

At the end, the temp venv would only have current version of brainsets installed, and any extra dataset-specific requirements.

### With these changes:

1. I am able to successfully start from a fresh environment and prepare the test dataset:
    ```bash
    python3.9 -m venv venv
    source venv/bin/activate
    pip install --upgrade pip
    pip install git+https://github.com/neuro-galaxy/brainsets@vinam/run_from_pip
    brainsets prepare pei_pandarinath_nlb_2021
    ```
2. `torch_brain`'s tests don't need to follow the weird requirement of running brainsets inside brainsets directory.
     (Working test @ https://github.com/neuro-galaxy/torch_brain/pull/70.)

### Problems with `uv run`
- Unfortunately I could get not it to layer dataset-specific requirements _on top of current environment_, when not inside brainsets directory. 
    - `uv run` follows different logic when its inside a python package directory (i.e. something that has `pyproject.toml`, `setup.py`, ...). This logic is what was helping with the layering happening correctly.
    - When not inside the brainsets package, it does not pass-on the installed brainsets package to the temp venv.
- Why not just add `brainsets` as a requirement of the dataset-specific `requirements.txt`? Because we would want to match local `brainsets` installation version with the version of the brainsets being installed in the temp venv.

---

A caveat: @ShivashriganeshMahato and I found that we need `brainsets` to be installed with `pip>=20.2` (happy to explain why in the comments if you're curious; it's a very nuanced problem that maybe 0.1% of users will see if they use `pip<20.2`). Which is why, I added `pip install --upgrade pip` to installation instructions in `README.md`.


brainsets/cli.py
```diff
@@ -31,7 +31,8 @@ def cli():
 @cli.command()
 @click.argument("dataset", type=click.Choice(DATASETS, case_sensitive=False))
 @click.option("-c", "--cores", default=4, help="Number of cores to use")
-def prepare(dataset, cores):
+@click.option("-v", "--verbose", is_flag=True, default=False)
+def prepare(dataset, cores, verbose):
```
vinamarora8, line 35: I used for debugging locally, but seems like this flag is good to keep for future debugs.

mazabou:
Ok, this looks quite complicated. I still need to take some time to review this, but I wanted to see what happens if we were using `pip install brainsets` from pypi
It doesn't work.

---
## Generating wheel (from main, not this branch)

I will start explaining how to generate the wheel and install it which replicates what happens with `pip install brainsets`:

first build the wheel (make sure you start with an empty env):
```
python3 -m pip install --upgrade build
python3 -m build
pip install dist/brainsets-0.1.1-py3-none-any.whl
```

this should install brainsets and you should see this:
```
>>> ls venv/lib/python3.9/site-packages | grep brainsets
brainsets
brainsets-0.1.1.dist-info
brainsets_pipelines
```

## First error: `brainsets` in site-pacakges vs. cloned `brainsets` 

now here's when it gets weird, after calling `brainsets config` do this:
```
brainsets prepare pei_pandarinath_nlb_2021
```
you get this error:
```
Preparing pei_pandarinath_nlb_2021...
Building temporary virtual environment using requirements from /Users/ma4766/Documents/Workspace/test_pip_for_brainsets/brainsets/venv/lib/python3.9/site-packages/brainsets_pipelines/pei_pandarinath_nlb_2021/requirements.txt
Uninstalled 1 package in 2ms
Installed 1 package in 6ms
error: File not found: `venv/lib/python3.9/site-packages/brainsets_pipelines/pei_pandarinath_nlb_2021/requirements.txt`
Error: Command failed with return code 2
```

now when you look inside of site-packages again, this is what you see:
```
>>> ls venv/lib/python3.9/site-packages | grep brainsets   
__editable__.brainsets-0.1.1.pth
__editable___brainsets_0_1_1_finder.py
brainsets-0.1.1.dist-info
```

somehow, `uv` makes a change. 1. `import brainsets; brainsets.__path__` points towards the cloned repo I used to make the build (before it had an independent copy of the code). and 2. there is no longer a `brainsets_pipelines` hence the error I get. 

not sure why `uv` does this.

if i am not inside of the `brainsets` repo clone, this won't happen. it seems like `uv` is somehow prioritizing the copy of brainsets in `.` instead of using the one in `site-packages`.

## Second error: `brainsets.taxonomy`

let's delete the brainsets installation, then do the same steps, but this time call `brainsets prepare pei_pandarinath_nlb_2021` in a different directory.

now this is the error i get:

It seems like `taxonomy.py` is not there, and that's what happens (this is the case even before `uv` is called).
```
ls /venv/lib/python3.9/site-packages/brainsets/     
__init__.py     cli.py          core.py         descriptions.py utils
```

## Third error: wrong numpy version.

It seems like uv is not resolving conflicts with already installed packages. it installs `h5py` with a newer numpy version for whatever reason, and ignores requirements from existing packages. 

```
/Users/ma4766/.cache/uv/archive-v0/uzDxqGDnUUK48zNgn3exK/bin/python3.9
Python 3.9.6 (default, Nov 11 2024, 03:15:38) 
[Clang 16.0.0 (clang-1600.0.26.6)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import brainsets
>>> brainsets.__path__
['/Users/ma4766/Documents/Workspace/test_pip_for_brainsets/brainsets/venv/lib/python3.9/site-packages/brainsets']
>>> import numpy
>>> numpy.__version__
'2.0.2'
>>> numpy.__path__
['/Users/ma4766/.cache/uv/archive-v0/uzDxqGDnUUK48zNgn3exK/lib/python3.9/site-packages/numpy']
>>> import scipy
/Users/ma4766/Documents/Workspace/test_pip_for_brainsets/brainsets/venv/lib/python3.9/site-packages/scipy/__init__.py:143: UserWarning: A NumPy version >=1.19.5 and <1.27.0 is required for this version of SciPy (detected version 2.0.2)
  warnings.warn(f"A NumPy version >={np_minversion} and <{np_maxversion}"
>>> scipy.__path__
['/Users/ma4766/Documents/Workspace/test_pip_for_brainsets/brainsets/venv/lib/python3.9/site-packages/scipy']
```

this is solved by adding `--with brainsets` which just installs a fresh copy of of brainsets, ensuring everything is good.

mazabou (changes requested):
I am worried that we are digging ourselves a bigger hole here. More code that needs to be tested in many environments...

I am tempted to just conclude that we need to add `brainsets` in the requirements file. 

We have the following users: Those who just want to download and prepare data: no issues here. Those who want to download brainsets to use it for their own prepare_data. 

in the latter case, they are forced to create a requirements.txt and specify the brainsets version they want to use (if we are using a cookiecutter this can be auto-populated). unfortunately, there could be a mismatch between their current brainsets in their environment, and the one they specified in requirements.txt. This leads to the conclusion that we should maybe allow the user to control their own environment (i.e. call `brainsets prepare` without a `requirements.txt` being forced on the user.) 

There is a third category of users: those who will make an edit of brainsets (which is a minority). in this case, having them be able to manage their own environment would be ideal anyways.

---

alternatively, the solution is to somehow have uv, make an exact copy of the environment, and figure out how it can respect the existing requirements, which is probably impossible to resolve? 

Let me know what you think.

vinamarora8:
I tried building a wheel from this branch, and it works for me:
```bash
cd ~/
mkdir brainsets_build_test
cd brainsets_build_test

# Clone brainsets (this branch)
git clone git@github.com:neuro-galaxy/brainsets.git
cd brainsets
git fetch origin
git checkout vinam/run_from_pip
python3.9 -m venv build_venv
source build_venv/bin/activate

# Build
pip install --upgrade pip
python3 -m pip install --upgrade build
python3 -m build
deactivate

# Install in a separate venv
cd ~/brainsets_build_test
python3.9 -m venv test_venv
source test_venv/bin/activate
pip install --upgrade pip
pip install brainsets/dist/brainsets-0.1.1-py3-none-any.whl

# Prepare data (assuming you have done `brainsets config`)
brainsets prepare pei_pandarinath_nlb_2021
```

The temp `venv` picks up the correct brainsets installation spec: `file:///home/varora38/brainsets_build_test/brainsets/dist/brainsets-0.1.1-py3-none-any.whl`

vinamarora8:
I am keen on having the temp venv have only two things installed:
- The `brainsets` installed in current `venv`
- Extra dataset-specific requirements
Because we would like the same `numpy`/`scipy` versions to be installed when preparing the data. 
This is because [numpy doesn't guarantee same RNG across different versions](https://numpy.org/neps/nep-0019-rng-policy.html#versioning)

vinamarora8:
I agree we will need to maintain this temporary `venv` solution. And people will find bugs that show up with other package managers (e.g. `pip<20.2` brings a bug where `pip freeze` is inaccurate).

But I am proposing that might be a worthy undertaking, because (i think) one important thing that `brainsets` would like to provide is reproducibility, i.e. same processed data no matter what machine it was processed on (within reason). This translates to us having a tight control over the environment that is used when processing the raw data.

I agree users who just do `pip install brainsets; brainsets config; brainsets prepare <dataset>` will not come close to any of the nuanced problems we're dealing here. 
But some users who would do `pip install git+https://...`, or the developers of `brainsets` who would do `pip install .`, would need what this PR is providing.

mazabou:
I think we are dealing with slightly orthogonal problems here.
The problem you are describing is inevitable, and cannot be solved by this PR. it requires specifying the version of each package. Simply think about using a package to process the data that eventually changes. but if in your requirements you just have "that_package", it will be auto updated to the latest version as determined by pip. this problem cannot be solved. the only thing we can do is monitor the data, and add the hashing system we discussed.

I can think of two things:
1. only allow merges to main that specify the numpy version in their requirements.txt (if they are using a random seed) -- this is obviously not enough but deals with a major source of randomness.
2. add the hash verification system, and have a "kernel" machine, that processes this data. if the data deviates it might signal a change in one of the packages that we can go investigate. The alternative is the process failing silently, and the users not even reporting an issue.

vinamarora8:
Closing this PR based on [discussion on Discord](https://discord.com/channels/1338561153089146962/1342996882313580594). 

Conclusions:
1. brainsets should be made part of `requirements.txt`.
2. Development mode will be handled by a flag `--use-active-env`.

PR #15, opened 2025-02-23
CLI improvements

vinamarora8:
## Program flow changes:
- Changes temp `venv` logic: Use `uv run * --isolated --no-project` instead of `uv run * --active`.
    This based on the decision to include `brainsets` in brainset `requirements.txt`.
- Removes the global `brainsets_pipelines/Snakefile`. 
   Instead the CLI will directly invoke the `all` rule in dataset specific Snakefiles.
- Restructures CLI code into `_cli/`.

### `brainsets prepare`:
- Add option `--raw-dir`: If provided, will override the value given in config 
- Add option `--processed-dir`: If provided, will override the value given in config
- Add option `--local`: To prepare a locally defined pipeline (path is given by the `dataset` argument). 
    Checks `<pipeline_dir>/Snakefile` exists before proceeding.
- Add flag `--use-active-env`: Bypasses temp `venv` creation completely. Sufficient warnings are added to deter normal users from using this flag.
- User is _not required_ to enter dataset name.
   If they simply do `brainsets prepare`, they are shown a list of available datasets, and will be prompted to enter the name _with autocompletion_ (try it!).
- Add flag `--verbose`/`-v`: For debugging purposes. If set, run both `snakemake` and `uv run` in verbose mode. Otherwise, run `snakemake` in `--quite` mode, and `uv run` in default mode.

### `brainsets config`:
- Change options `--raw`/`--processed` to `--raw-dir`/`--processed-dir` to stay consistent with `brainsets prepare`.

---

Working with pip install outside this repository: https://github.com/neuro-galaxy/torch_brain/pull/70


brainsets_pipelines/perich_miller_population_2018/requirements.txt
```diff
@@ -1 +1,2 @@
-dandi==0.61.2
\ No newline at end of file
+dandi==0.61.2
+brainsets @ git+https://github.com/neuro-galaxy/brainsets
```
vinamarora8, line 2: Before merging, make sure to set the brainsets version to an appropriate one.

brainsets_pipelines/pei_pandarinath_nlb_2021/requirements.txt
```diff
@@ -1 +1,2 @@
-dandi==0.61.2
\ No newline at end of file
+dandi==0.61.2
+brainsets @ git+https://github.com/neuro-galaxy/brainsets
```
vinamarora8, line 2: Before merging, make sure to set the brainsets version to an appropriate one.

brainsets_pipelines/pei_pandarinath_nlb_2021/requirements.txt
```diff
@@ -1 +1,2 @@
-dandi==0.61.2
\ No newline at end of file
+dandi==0.61.2
+brainsets @ git+https://github.com/neuro-galaxy/brainsets@vinam/cli2
```
vinamarora8, line 2: Before merging, make sure to set the brainsets version to an appropriate one.

brainsets_pipelines/perich_miller_population_2018/requirements.txt
```diff
@@ -1 +1,2 @@
-dandi==0.61.2
\ No newline at end of file
+dandi==0.61.2
+brainsets @ git+https://github.com/neuro-galaxy/brainsets@vinam/cli2
```
vinamarora8, line 2: Before merging, make sure to set the brainsets version to an appropriate one.

brainsets/_cli/cli_list.py
```diff
@@ -0,0 +1,11 @@
+import click
+
+from .utils import get_available_datasets
+
+
+@click.command()
+def cli_list():
+    """List available datasets."""
+    click.echo("Available datasets:")
```
mazabou, line 9: do we want to stick with "brainsets" instead of "datasets"?

brainsets/_cli/utils.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def save_config(config):
+    with open(CONFIG_FILE, "w") as f:
+        yaml.safe_dump(config, f, default_flow_style=False)
+    return CONFIG_FILE
+
+
+def get_available_datasets():
+    return [d.name for d in PIPELINES_PATH.iterdir() if d.is_dir()]
```
mazabou, line 60: when I run `brainsets list`, I get the following:
```
$ brainsets list   
Available datasets:
- pei_pandarinath_nlb_2021
- .snakemake
- ibl_reproducible_ephys_2022
- perich_miller_population_2018
```

it's possibly because I ran something inside of the PIPELINE_PATH. not sure how we would want to deal with this? Should we only return "valid" pipelines?

brainsets/_cli/cli_prepare.py
```diff
@@ -0,0 +1,175 @@
...
+        # Prompt user if dataset is not provided
+        if dataset is None:
+            click.echo(f"Available datasets: ")
+            for dataset in available_datasets:
+                click.echo(f"- {dataset}")
+            click.echo()
+            dataset = prompt(
+                "Enter dataset name: ",
+                auto_suggest=AutoSuggestFromList(available_datasets),
+            )
```
mazabou, line 91: is there a way of doing the auto suggest at the level of brainsets prepare ? i.e. when you are calling it, and then pressing tab for example. I don't know about allowing `brainsets prepare` without any argument.

brainsets/_cli/cli_prepare.py
```diff
@@ -0,0 +1,175 @@
...
+        # Check dataset name validity
+        if dataset not in available_datasets:
+            raise click.ClickException(f"Invalid dataset name: {dataset}")
```
mazabou, line 94: maybe `"brainset {dataset} not found. Run 'brainsets list' to get the available list of datasets`

brainsets/_cli/cli_prepare.py
```diff
@@ -0,0 +1,175 @@
...
+        # Find snakefile
+        snakefile_filepath = PIPELINES_PATH / dataset / "Snakefile"
+        reqs_filepath = PIPELINES_PATH / dataset / "requirements.txt"
+        click.echo(f"Preparing {dataset}...")
+
+    # If local pipeline directory provided
+    else:
+        pipeline_dir = expand_path(pipeline_dir)
```
mazabou, line 102: I liked the way you originally pitched this, for the public brainsets, you always have to run:
```
brainsets prepare <brainset_id>
```
and locally you can alternatively run:
```
brainsets prepare ./path/to/pipeline/custom_brainset_id
```

currently `brainsets prepare` has too many options, in some case you will be asked for a brainset_id and in other you won't?

mazabou: alternatively forcing the user to still enter a unique brainset_id like so:

```
brainsets prepare my_brainset_id --pipeline-path ./path/to/pipeline/custom_brainset_id
```

this way:
1. you can check that this will not overwrite an existing folder
2. you can keep the api a bit more consistent, since you would always have to input a brainset_id as the positional argument
3. you can add this to the brainsets list ?

mazabou (changes requested):
Overall pretty major upgrade! I really like the autocomplete as well.

Left a few edit suggestions/ questions. 

Did you give up on listing local pipelines when calling `brainsets list`?

brainsets/_cli/cli_prepare.py
mazabou: also i have to use the right arrow instead of using tab currently. can that be fixed?

vinamarora8:
> Did you give up on listing local pipelines when calling brainsets list?
Yes, holding that off for now. `brainsets prepare` allowing local pipelines covers my use case.

brainsets/_cli/utils.py
vinamarora8: Woops, that's a bug. I'll fix it.

brainsets/_cli/cli_prepare.py
vinamarora8: > is there a way of doing the auto suggest at the level of brainsets prepare ? i.e. when you are calling it, and then pressing tab for example.

We have to ask users to add a "shell completion" function to their shell's dot-rc. It's possible only then. Not a fan of that.

> I don't know about allowing brainsets prepare without any argument.

What's the problem?
I think it's good. Dataset names are complicated and I saw myself following this 90% of the time: First `brainsets list`, then copy the dataset name, and then `brainsets prepare <dataset>`. This will save people the three steps.

> also i have to use the right arrow instead of using tab currently. can that be fixed?

I would also prefer using tab, but I have found that most cmd-line autocompletes follow right-arrow for selecting an _in-line_ completion suggestion (see `ipython`, `zsh`, ...). Tab-completion is standard when there is a drop-down of available completions (`prompt_toolkit` allows drop-down suggestions, but in-line completion looks cleaner). I think it's just muscle memory that we're not familiar with yet. People who use `zsh` (again, I don't) use right-arrow when they see in-line completion suggestion all the time. So I say we stick to that established thing.

brainsets/_cli/cli_prepare.py
vinamarora8: It's simple. The brainset-id argument is only an identifier to a predefined pipeline's path. The _real_ brainset id is defined inside `prepare_data.py`. If you are using a local pipeline path, then specifying brainset-id in the command line has no job.

Problem with 
> `brainsets prepare ./path/to/pipeline/custom_brainset_id`

What if I do
```bash
brainsets prepare dataset_xyz
```
`dataset_xyz` is a valid predefined pipeline, and I also have a local folder in the current directory called `dataset_xyz` that is a valid pipeline dir. Should we choose the local pipeline or predefined one?
It seems cleaner to make it explicit when the user is asking for local pipeline and when they're asking for a predefined one. This is a power user feature, so expecting power users to know one more option (or do `brainsets prepare --help`) is fair.

> you can add this to the brainsets list ?

I know I originally proposed showing local datasets in `brainsets list`. But now I think it's too-much-feature-for-too-little-gain. Right now, there is a clean separation:
- If you don't use `--pipeline-dir`, then you're asking for a predefined pipeline.
- If you use `--pipeline-dir`, then you're asking for local pipeline and will provide the path.

mazabou: > dataset_xyz is a valid predefined pipeline, and I also have a local folder in the current directory called dataset_xyz that is a valid pipeline dir. Should we choose the local pipeline or predefined one?
> It seems cleaner to make it explicit when the user is asking for local pipeline and when they're asking for a predefined one. This is a power user feature, so expecting power users to know one more option (or do brainsets prepare --help) is fair.
> 

this is the exact problem we want to avoid, and hence why we want to enforce the use of brainset_id, so that this doesn't happen. if you have a local version of some pipeline, you should name it something different, to avoid confusion anyways.

mazabou: if you want to make a clear separation between og brainsets, and local ones, you can use different cli commands. there is a total of 5 arguments, and 2 of them are exclusive (brainset_id, and pipeline-dir)

brainsets/_cli/cli_prepare.py
mazabou: I still would like to have the autocomplete happen at the level of brainsets prepare. alternatively, we can come up with another command that doesn't feel incomplete when you are just calling `brainsets prepare` 
it's like doing `git add` and then being prompted for files later. even in the description you are suggesting the user to do `brainsets prepare <DATASET>`. 

---

I am pretty certain that the most common way of accepting the auto complete is "tab", I just spent 10 min googling this. I am happy to enable both?

brainsets/_cli/cli_prepare.py
mazabou: what i am saying is that if you want to allow brainsets prepare --pipeline-dir, maybe use something different than prepare

vinamarora8: @mazabou bump

vinamarora8:
@mazabou Ready for another review.

Changes since last discussion:
- Fix bug where ".snakemake" appeared in `brainsets list`: Ignore any directory names that start with "." or "_"
- Remove autcomplete from `brainsets prepare`
- Change `brainsets prepare --pipeline-dir` to `brainsets prepare <path-to-local-pipeline> --local`
- Improve validation of the Snakefile before proceeding (check if Snakefile exists, and has the "all" rule defined)

vinamarora8:
Additional change:
- `brainsets config` supports tab-based autocompletion for the raw and processed directories.

brainsets/_cli/cli_prepare.py
```diff
@@ -0,0 +1,201 @@
...
+    local: bool,
+):
+    """Download and process a single brainset.
+
+    Run 'brainsets list' to get a list of available brainsets.
+
+    \b
+    Examples:
+    $ brainsets prepare pei_pandarinath_nlb_2019
+    $ brainsets prepare pei_pandarinath_nlb_2019 --cores 8 --raw-dir ~/data/raw --processed-dir ~/data/processed
```
mazabou, line 69: this should say `pei_pandarinath_nlb_2021`

mazabou (approved):
This is awesome thanks @vinamarora8! I downloaded and tested the different commands, works great.

I found an error in the description, but otherwise, looks good to me, let's merge!

brainsets/_cli/cli_prepare.py
vinamarora8: woops

vinamarora8: Can you fix this? You have to edit the brainsets versions in requirements.txt too

PR #18, opened 2025-02-26
update perich_miller_population_2018

mazabou:


ximmao:
Should there be cursor_vel normalization? previous one has `cursor_vel = cursor_vel / 20.0`, I haven't found it here.

mazabou:
> Should there be cursor_vel normalization? previous one has `cursor_vel = cursor_vel / 20.0`, I haven't found it here.

Normalization is now done in the configuration file when training.

PR #19, opened 2025-03-03
Pass `PIPELINE_DIR` to Snakefile

vinamarora8:
Currently, Snakefile only receives `RAW_DIR` and `PROCESSED_DIR` as inputs from the command line. 

For local pipelines I would like a `PIPELINE_DIR` to be passed to the Snakefile, so that it can correctly call the script `PIPELINE_DIR/prepare_data.py`. Without the knowledge of pipeline path, a local Snakefile has no knowledge of where the `prepare_data` script is located.

Additional edits:
- Adjusted current Snakefiles to also follow this format.
- Cleared up some code repetitions within both the Snakefiles.


brainsets_pipelines/pei_pandarinath_nlb_2021/Snakefile
```diff
@@ -5,48 +5,49 @@
 DATASET = "pei_pandarinath_nlb_2021"
 DANDI_ID = "000140"
 
-RAW_DIR = config["RAW_DIR"]
-PROCESSED_DIR = config["PROCESSED_DIR"]
+RAW_DIR = Path(config["RAW_DIR"]) / DATASET
+PROCESSED_DIR = Path(config["PROCESSED_DIR"]) / DATASET
```
mazabou, line 9: Not a big fan of this, config["RAW_DIR"] would have different meanings. also doesn't seem to be relevant to this PR?

mazabou (commented):
Could you explain what you are trying to do here? If you are editing a local copy of brainsets then you can simply install it with `pip install -e .` and you wouldn't need to do any of this? That's what I have been doing.

If you are using a custom dataset, then you are specifying the path to the pipelines any ways. I am a bit confused of how this would be used.

vinamarora8:
For future ref: I was talking about the workflow where you don't clone `brainsets`, but do `pip install brainsets`; and want to run a local pipeline.

We found `workflow.base_dir` inside Snakemake can give us the `PIPELINE_DIR`, so this feature is unneeded and the PR can be closed.

PR #39, opened 2025-11-08
Transfer and test pipeline for removing snakefiles 

AlexandreAndr:


brainsets_pipelines/churchland_shenoy_neural_2012/prepare_data.py
```diff
@@ -53,15 +199,15 @@ def extract_trials(nwbfile, session_id):
     )[0]
     artifact_dict = {}
     if len(artifact_index) > 0:
-        logging.info(f"Found {len(artifact_index)} artifacts in the trial table.")
+        # logging.info(f"Found {len(artifact_index)} artifacts in the trial table.")
```
ShivashriganeshMahato: Note so we don't forget: we should have logging support and re-add these statements before merging.

PR #43, opened 2025-11-11
Migrate odoherty_sabes_nonhuman_2017 to new pipeline

ShivashriganeshMahato:
Successfully downloads and processes odoherty_sabes_nonhuman_2017. The pipeline validates md5 checksums after download, and uses it to avoid redownloading on subsequent runs (respecting `--redownload` and `--reprocess`, tested both). On two runs with fresh data directories, downloading + preprocessing took a total of 6m3s & 7m44s. For reprocessing it took 28s and for redownloading it took 5m56s.


vinamarora8 (approved):
lgtm

PR #54, opened 2025-11-28
Adding SleepEDF to Brainsets

milosobral:
Adding this EEG dataset commonly used for sleep-staging: https://www.physionet.org/content/sleep-edfx/1.0.0/

Using the aws interface for creating the manifest and download as it was much faster than wget.

### DIFF
--- a/brainsets_pipelines/kemp_sleep_edf_2013/pipeline.py
+++ b/brainsets_pipelines/kemp_sleep_edf_2013/pipeline.py
@@ -1,3 +1,12 @@
+# /// brainset-pipeline
+# python-version = "3.11"
+# dependencies = [
+#   "mne~=1.11.0",
+#   "boto3~=1.41.0",
+#   "scikit-learn==1.7.2",
+# ]
+# ///
+
 from argparse import ArgumentParser
 from pathlib import Path
 from typing import Optional, Tuple
--- a/brainsets_pipelines/kemp_sleep_edf_2013/requirements.txt
+++ b/brainsets_pipelines/kemp_sleep_edf_2013/requirements.txt
@@ -1,3 +1,0 @@
-mne~=1.11.0
-boto3~=1.41.0
-scikit-learn==1.7.2


brainsets_pipelines/kemp_sleep_edf_2013/pipeline.py
```diff
@@ -0,0 +1,448 @@
...
+
+        data_dict = {
+            "brainset": brainset_description,
+            "subject": subject,
+            "session": session_description,
+            "device": device_description,
+            "eeg": signals,
+            "units": units,
+            "domain": signals.domain,
+        }
```
estefanysuarez: Very nitpicky, but why not follow the convention of adding brainset, subject, session, device, eeg and units directly into the Data object (instead of everything into a Dict())? Only enclose behavioural variables into a Dict() and then unpack it

brainsets_pipelines/kemp_sleep_edf_2013/notebook_to_test.ipynb
vinamarora8: Accidental file add?

brainsets/utils/split.py
```diff
@@ -168,3 +169,190 @@ def generate_train_valid_test_splits(epoch_dict, grid):
...
+    if hasattr(intervals, "keys"):
+        for key in intervals.keys():
+            if key in ["start", "end"]:
+                continue
+            val = getattr(intervals, key)
+            if isinstance(val, (np.ndarray, list)) and len(val) == len(intervals):
+                if isinstance(val, np.ndarray):
+                    kwargs[key] = val[original_indices]
+                else:
+                    kwargs[key] = [val[i] for i in original_indices]
```
vinamarora8: What's a value with type `list`? I think everything pubic in `Interval` should be a `np.array`?

brainsets/utils/split.py
```diff
@@ -168,3 +169,190 @@ def generate_train_valid_test_splits(epoch_dict, grid):
...
+        ValueError: If the intervals don't have the specified attribute.
+    """
+    if not hasattr(intervals, stratify_by):
+        raise ValueError(
+            f"Intervals must have a '{stratify_by}' attribute for filtering."
+        )
+
+    attr_values = getattr(intervals, stratify_by)
+    mask = ~np.isin(attr_values, exclude_values)
+    return intervals.select_by_mask(mask)
```
vinamarora8: This whole function seems to be replaceable with a simple one-liner.

```python
x.select_by_mask(np.isin(x.id, exclude_values, invert=True))
```

You sure you want to have this as a function? 

If so, we should follow API signature of another library that has this (idk if numpy/pandas have this). We should rename parameter `stratify_by` to something more reasonable.

brainsets/utils/split.py
```diff
@@ -168,3 +169,190 @@ def generate_train_valid_test_splits(epoch_dict, grid):
...
+
+
+def _create_interval_split(intervals: Interval, indices: np.ndarray) -> Interval:
+    """Create an Interval subset from indices and sort it."""
+    mask = np.zeros(len(intervals), dtype=bool)
+    mask[indices] = True
+    split = intervals.select_by_mask(mask)
+    split.sort()
+    return split
```
vinamarora8: Note for future: We should make this into a method of `ArrayDict`, etc.

brainsets_pipelines/kemp_sleep_edf_2013/pipeline.py
```diff
@@ -0,0 +1,404 @@
...
+            prefix = base_name[:prefix_len]
+            search_prefix = str(psg_path.parent / prefix)
+
+            response = s3.list_objects_v2(Bucket=BUCKET, Prefix=search_prefix)
+
+            for obj in response.get("Contents", []):
+                if "Hypnogram" in obj["Key"] and obj["Key"].endswith(".edf"):
+                    return obj["Key"]
+
+        return None
```
vinamarora8: Maybe nit: Instead of having this as a `classmethod`, you could define this as a function instead the `get_manifest` classmethod?

vinamarora8 (changes requested):
Awesome work Milo! I have reviewed `utils/splits.py` today, and will review the actual pipeline soon. Left comments below:

brainsets_pipelines/kemp_sleep_edf_2013/pipeline.py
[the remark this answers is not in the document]
vinamarora8: +1 for putting Pipeline class definition at the beginning (i.e. before `find_hypnogram_key`)

pyproject.toml
```diff
@@ -32,11 +32,20 @@ dependencies = [
     "prompt_toolkit",
     "numpy>=1.14.0",
     "pulp<2.8",
-    "ray>=2.30.0"
+    "ray>=2.30.0",
+    "h5py>=3.14.0",
+    "mne>=1.8.0",
+    "scikit-learn>=1.0.0",
```
vinamarora8: I don't think we _need_ to add these two at a package level. Any problems if we let these be strictly at the pipeline level? See my comment at the start of `utils/split.py`

brainsets_pipelines/kemp_sleep_edf_2013/requirements.txt
```diff
@@ -0,0 +1,3 @@
+mne~=1.11.0
+boto3~=1.41.0
+scikit-learn~=1.7.2
```
vinamarora8: Curious as to why not `==`?

brainsets_pipelines/kemp_sleep_edf_2013/pipeline.py
```diff
@@ -0,0 +1,424 @@
...
+        elif "eog" in ch_name_lower:
+            modality = Modality.EOG
+        elif "emg" in ch_name_lower:
+            modality = Modality.EMG
+        elif "resp" in ch_name_lower:
+            modality = Modality.RESP
+        elif "temp" in ch_name_lower:
+            modality = Modality.TEMP
+        else:
+            continue
```
vinamarora8: Nit: Defining an Enum for just this seems a bit overkill

vinamarora8 (changes requested):
Reviewed the pipeline and leaving some more comments

brainsets/utils/s3_utils.py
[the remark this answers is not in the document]
milosobral: I tend to agree but all other files follow this same convention (`dandi_utils.py`, `dir_utils.py`, `mat_utils.py`)

brainsets/utils/s3_utils.py
```diff
@@ -0,0 +1,28 @@
+import boto3
+from botocore import UNSIGNED
+from botocore.config import Config
+
+
+def get_s3_client(
+    max_retries: int = 5,
+    max_pool_connections: int = 10,
+) -> boto3.client:
```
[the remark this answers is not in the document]
milosobral: I think 99% of the time, any S3 client will be used like this. I can rename it get_s3_client_for_download maybe?

brainsets/utils/split.py
vinamarora8: I guess now this can be replaced with `Interval.segment` + an overlap check? With the tests for this also removed.

brainsets_pipelines/kemp_sleep_edf_2013/pipeline.py
```diff
@@ -0,0 +1,402 @@
...
+    subject_info = info.get("subject_info", {})
+
+    age = subject_info.get("age")
+    try:
+        age_str = subject_info.get("last_name")
+        if age_str is not None:
+            age = int(age_str.replace("yr", ""))
+    except (ValueError, AttributeError) as e:
+        logging.warning(f"Could not parse age from last_name: {age_str}")
+        age = None
```
vinamarora8: This still remains a little confusing block to read. Can you add some comments?

What will `age = subject_info.get("age")` be like? Why are you overwriting with something from"last_name", and why overwrite with `None` when something goes wrong?

vinamarora8 (changes requested):
Changes look good overall. Leaving some minor comments

brainsets/utils/split.py
vinamarora8: Aah I just realized that's only once we release a new temporaldata version. So, I guess, having this here makes sense.

brainsets_pipelines/kemp_sleep_edf_2013/pipeline.py
milosobral: that was old code I forgot to remove good catch

vinamarora8 (approved):
looks good! ready to merge

PR #57, opened 2025-12-03
SleepEDF New Splits Resolver

milosobral:
Example of what the new split resolver could look like

Depends on #54


milosobral:
This functionality will be included with the new dataset proposal in https://github.com/neuro-galaxy/torch_brain/pull/154

PR #58, opened 2025-12-05
Adding OpenNeuro base pipeline

oumgharbi:
Adding a base pipeline for downloading and processing OpenNeuro EEG datasets. 

The module provides default implementations to enumerate available sessions, download each session then process it into `temporaldata.Data` objects.

OpenNeuro datasets can then be added using this base module. 
The OpenNeuro dataset directory is structured as:
````
{my_brainset_name}/
├── {dataset_id}_config.json
├── pipeline.py
└── requirements.txt
````
For each dataset, a configuration file should be provided to describe all metadata (dataset id, participants, channels, recording info, etc.), which will be used to process sessions. For now, an LLM was used to generate these metadata.

The dataset-specific pipeline typically relies on the base pipeline module for default implementations, and can add specific processing steps if needed.

The [The Bitbrain Open Access Sleep (BOAS) dataset](https://openneuro.org/datasets/ds005555/versions/1.0.0) is added as a first example under [`brainsets_pipelines/klinzing_sleep_ds005555_2024`](https://github.com/neuro-galaxy/brainsets/tree/oumayma/openneuro/brainsets_pipelines/klinzing_sleep_ds005555_2024).

### DIFF
--- a/tests/test_open_neuro_pipeline.py
+++ b/tests/test_open_neuro_pipeline.py
@@ -248,26 +248,32 @@ class TestDownload:
         item = pd.Series(
             {
                 "dataset_id": "ds006695",
-                "s3_url": "s3://openneuro.org/ds006695/sub-1_task-Sleep_acq-headband",
+                "version_tag": "1.0.0",
                 "subject_id": "sub-1",
                 "subject_info": {"age": "25", "sex": "F"},
+                "task_id": "Sleep",
+                "channel_mapping": {"Fp1": ("Fp1", "EEG")},
                 "device_info": {
                     "device_name": "Test Device",
                     "device_manufacturer": "Test Manufacturer",
                 },
-                "channel_mapping": {"Fp1": ("Fp1", "EEG")},
+                "s3_url": "s3://openneuro.org/ds006695/sub-1_task-Sleep_acq-headband",
+                "fpath": Path("/tmp/raw/sub-1"),
             },
             name="sub-1_task-Sleep_acq-headband",
         )
 
         mock_item = MagicMock()
         mock_item.dataset_id = item["dataset_id"]
+        mock_item.version_tag = item["version_tag"]
         mock_item.s3_url = item["s3_url"]
         mock_item.Index = item.name
         mock_item.subject_id = item["subject_id"]
         mock_item.subject_info = item["subject_info"]
+        mock_item.task_id = item["task_id"]
         mock_item.device_info = item["device_info"]
         mock_item.channel_mapping = item["channel_mapping"]
+        mock_item.fpath = item["fpath"]
         return mock_item
 
     @patch("brainsets.utils.open_neuro_pipeline.check_recording_files_exist")

### CODE REVIEW

File: brainsets/utils/open_neuro_pipeline.py [hunk 1]
```diff
@@ -0,0 +1,526 @@
...
+
+        This default implementation handles common OpenNeuro EEG processing tasks:
+        1. Finds EEG files in BIDS structure
+        2. Loads them using MNE
+        3. Extracts metadata (subject, session, device, brainset descriptions)
+        4. Extracts EEG signal and channel information
+        5. Creates a Data object and saves it to disk
+
+        Subclasses can override this method and call super().process() to use
+        the default implementation, then add dataset-specific processing logic.
```
milosobral: With this implementation, adding dataset-specific processing logic would require reloading and resaving the Data object to and from disk. What if instead, you wrap all the common implementation (except for the saving of the object) into a private function and call that from `process()`. When someone wants to overwrite the behavior, they don't call `super.process` but instead take advantage of tha wrapper function and then add their behavior. 

Something like this: 

```
def _process_common(self, download_output: dict):
    # Do all the stuff for processing
    data = Data(...)
    return data

def process(self, download_output: dict):
    data = self._process_common(download_output: dict)
    with h5py ...
        data.to_hd5f(...)

# For dataset specific Pipeline
def process(self, download_output: dict):
    data = self._process_common(download_output: dict)

    # ADD DATASET SPECIFIC CODE HERE

    with h5py ...
        data.to_hd5f(...)
```

brainsets_pipelines/klinzing_sleep_ds005555_2024/pipeline.py
```diff
@@ -0,0 +1,58 @@
...
+    - :meth:`process()` – loads EEG with MNE, extracts metadata and signal,
+      and stores a standardized :class:`temporaldata.Data` object.
+
+    Dataset-specific processing can be added later by overriding :meth:`process()`
+    and calling ``super().process(download_output)``.
+    """
+
+    # Required identifiers
+    brainset_id = "klinzing_sleep_ds005555_2024"
+    dataset_id = "ds005555"
```
milosobral: Could you read those directly from the config file? This avoids some copy pasting and would limit the possibility of this variable and what is in the config file to become unsynchronized for some reason

milosobral (commented):
Looks really good! Added a few comments but didn't look too much into the Processing and haven't actually tried it yet as you said you were making some modifs to processing.

brainsets/utils/open_neuro_pipeline.py
```diff
@@ -0,0 +1,526 @@
...
+                - 's3_url': S3 URL for downloading
+                - 'fpath': Local file path where the manifest item will be downloaded
+            The index is set to 'recording_id'.
+        """
+        dataset_id = validate_dataset_id(cls.get_dataset_id())
+
+        if cls.version_tag is None:
+            version_tag = fetch_latest_version_tag(dataset_id)
+        else:
+            version_tag = cls.version_tag
```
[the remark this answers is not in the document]
oumgharbi: So for now, we can set the version tag according to the config file to be synchronous. In case of discrepancies, we can fix the config file later.

brainsets/utils/open_neuro_pipeline.py
```diff
@@ -0,0 +1,599 @@
...
+        )
+
+        if not recording_info:
+            raise ValueError(
+                f"No recording_info found in config file. "
+                f"Expected path: dataset.recording_info.recording_info"
+            )
+
+        # Filter recordings if subject_ids is specified
+        # FIXME
```
milosobral: Is something here not working yet?

brainsets/utils/open_neuro_pipeline.py
```diff
@@ -0,0 +1,599 @@
...
+
+            subject_id = rec.get("subject_id")
+            subject_info = rec.get("participant_info", {})
+            fpath = raw_dir / subject_id / recording_id if subject_id else None
+            # FIXME fpath = raw_dir / subject_id / eeg /recording_id
+
+            channel_map_id = rec.get("channel_map_id")
+            channel_mapping = None
+            if channel_map_id and channel_map_id in channel_maps:
+                channel_map = channel_maps[channel_map_id].get("channels", {})
```
milosobral: Same here, wouldn't we need to assume that we do have the channel in the channel maps?

brainsets/utils/open_neuro_pipeline.py
```diff
@@ -0,0 +1,599 @@
...
+                channel_mapping = {}
+                device_info = {}
+                for old_name, channel_info in channel_map.items():
+                    if isinstance(channel_info, dict) and "new_name" in channel_info:
+                        new_name = channel_info["new_name"]
+                        modality = channel_info.get("modality", "UNKNOWN")
+                        channel_mapping[old_name] = (new_name, modality)
```
milosobral: I'm not sure I understand why iterating through the channel map is necessary here. Couldn't you just pass the channel map directly?

brainsets/utils/open_neuro_pipeline.py
```diff
@@ -0,0 +1,599 @@
...
+    @abstractmethod
+    def _get_identifiers(cls) -> tuple[str, str]:
+        """Get brainset_id and dataset_id for this pipeline.
+
+        This method must be implemented by subclasses. It can either:
+        - Load identifiers from a config file
+        - Return hardcoded values
+
+        """
+        ...
```
milosobral: Not sure I understand why the implementation of this is left to the subclasses? Couldn't we simply assume that all of the read from the Config file?

milosobral (commented):
Great work! 

Added some questions where things needed clarification as well as some implementations that we might want to simplify/make more clear.

Otherwise, it seems like you have left a lot of #TODO or #FIXME comments, are those for things where you weren't sure of what to do? It seems like some of them can directly be adressed or removed.

PR #59, opened 2025-12-11
Added single session download util

milosobral:


milosobral:
> I'm not sure what's the advantage over doing this in the CI:

You're right, this is probably some added complication for no real added benefits.

Issue #52, opened 2025-11-22
Python version in individual pipelines


vinamarora8:
## Problem
Currently, some brainset pipelines implicitly depend on very specific python versions; and do not work for all the python versions supported by the `brainsets` package. 

For instance the following succeeds when run with py3.9, but fails with run with py3.10+ 
```bash
uv venv venv --python=$PYTHON_VERSION
source venv/bin/activate
uv pip install brainsets
brainsets prepare perich_miller_population_2018 --raw-dir ./raw --processed-dir ./processed
```
This happens because `dandi_cli` behaves differently with py3.9 and py3.10+. While this is a specific example, this is probably already is an issue for multiple pipelines.

## Solution
We should allow pipeline authors to specify which python version should be used to run their pipeline.
Once we have this, the CLI can simply ask `uv run` to the pipeline with a specific requirement.

The only decision to be made is _how_ should pipelines indicate the python version spec.

vinamarora8:
Re: how should pipelines indicate which python version

**Proposal 1:** Adapt `uv run`'s approach for dependency and python version declaration. 
In this, the following will go at the top of the `prepare_data.py` files:
```python
# /// brainset_pipeline
# requires-python = "==3.9"
# dependencies = [
#   "dandi==0.61.2",
#   "scikit-learn==1.2.1",
# ]
# ///

# rest of the code for the pipeline
```
Our CLI can read this declaration from the top and parse it for both - python requirement, and dependency list.  We will have to write our own parser for the metadata block, but that seems trivial. This allows us to have a single file defining everything about the pipeline - including dependencies and python requirement!

The above codeblock is adapted from examples in uv's doc: [link](https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies).

@milosobral @mazabou, what are your thoughts?

PR #68, opened 2025-12-23
v0.2.0

vinamarora8:
I think things have change enough to warrant an increment of the minor version counter. Happy to discuss further!

This PR:
1. Adds tag-based versioning and CI/CD pipeline similar to neuro-galaxy/torch_brain#146. 
    Deviations from that PR:
    - Only publish to PyPI since we have not decided what to use TestPyPI for. We can add this later.
    - Pre-release versions will also go to PyPI (this, I learned recently, is the standard)
    - Adds validation for version syntax in the publish action
2. Adds `RELEASE.md` formalizing the release process and version syntax.
3. Updates `CHANGELOG.md` for next release
4. Updates `README.md` with new brainsets
5. Updates dependencies: CI testing on `Py3.9` are mysteriously breaking due to dropped `Py3.9` support in newer versions of `ray`. For some reason, dependency resolver is unable to solve this ([see failed run](https://github.com/neuro-galaxy/brainsets/actions/runs/20450518669/job/58762662683)). Dependencies in `pyproject.toml` have been updated to help the resolver.

## Summary by CodeRabbit

* **Chores**
  * Added an automated CI/CD pipeline to publish releases to PyPI.
  * Enabled dynamic package versioning and updated build configuration and packaging requirements.

* **Documentation**
  * Added a release guide documenting versioning conventions and release steps.
  * Added an Unreleased section to the changelog.

* **Tests**
  * Added a test to verify the package exposes a non-empty version attribute.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -3,7 +3,8 @@
 1. Ensure `main` is ready to release.
 2. Make sure `CHANGELOG.md` is up to date (review commit history to verify, open a Release PR if changes are needed).
 3. Verify the version in `CHANGELOG.md` matches the version you're about to tag.
-4. Create and push the release tag ([see here](#valid-version-numbers-and-their-meaning) for version syntax).
+4. If a new brainset pipeline has been added, verify that it is mentioned in the project README.
+5. Create and push the release tag ([see here](#valid-version-numbers-and-their-meaning) for version syntax).
 
 ```bash
 $ git checkout main
@@ -12,8 +13,8 @@ $ git tag vX.Y.Z
 $ git push origin vX.Y.Z
 ```
 
-5. Monitor the GitHub Actions workflow to ensure the build and publish succeed.
-6. Add a GitHub Release with a small description of changes from the last release.
+6. Monitor the GitHub Actions workflow to ensure the build and publish succeed.
+7. Add a GitHub Release with a small description of changes from the last release.
 
 **Note:** Tags must start with "v" for the publishing GitHub Action to begin.
 


CHANGELOG.md
```diff
@@ -3,6 +3,26 @@
...
+### Removed
+
+### Changed
+
+## [0.2.0] - 2025-12-24
+### Added
+- New brainset pipeline: `allen_visual_coding_ophys_2016` ([#16](https://github.com/neuro-galaxy/brainsets/pull/16)).
+- Added `STEREO_EEG` to `RecordingTech` enum ([#36](https://github.com/neuro-galaxy/brainsets/pull/36)).
+- Added `BrainsetPipeline` definition with a `brainset.runner` module to run `BrainsetPipeline` in parallel ([#37](https://github.com/neuro-galaxy/brainsets/pull/37)).
+- Added support for PEP723-style inline metadata blocks to specify Python version and dependencies in a `pipeline.py` file ([#62](https://github.com/neuro-galaxy/brainsets/pull/62)).
```
mazabou: This PR adds a level of complexity that i am not a big fan of. The issue that is referenced can be solved by adding a `.python-version` file (see https://pydevtools.com/handbook/explanation/what-is-a-python-version-file/), instead a new (non standard) way of defining the environment is introduced. 

I can be convinced that we should do it this way, so can you explain what problem is being solved here?

mazabou (changes requested):
I took a look at most PRs. There is only one that I believe is a bit overly complicated, and I would suggest holding off on it for a future release until we discuss it more.

CHANGELOG.md
vinamarora8: > so can you explain what problem is being solved here?

Please have a look at that PR's attached issue https://github.com/neuro-galaxy/brainsets/issues/52 to understand the problem, the proposal, and its underlying reasoning.

> The issue that is referenced can be solved by adding a .python-version

While you may be more familiar with `.python-version`, the PEP-723 metadata-based approach is a new Python standard and is what newer tools like `uv` are adopting ([link](https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies)). It is based off of [PEP-723](https://peps.python.org/pep-0723/), and so I would disagree with calling it "non-standard".

At the end of the day, both approaches are inspired by "standards" created by other tools, and which to choose is almost a matter of opinion. There is, however, one benefit of the PEP-723 style metadata approach: If we want to add and support more metadata fields in the future, we can easily add them without having users create even more files.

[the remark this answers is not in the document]
vinamarora8: These Classes and Modules are new additions. They provide features that were previously absent, and so I believe they should remain in Added. There is a corresponding entry in "Changed" which expresses your sentiment that the pipelines have changed:

> - Updated all existing pipelines with `BrainsetPipeline` and dependencies in metadata block ([#37](https://github.com/neuro-galaxy/brainsets/pull/37) and [#62](https://github.com/neuro-galaxy/brainsets/pull/62)).

PR #74, opened 2026-01-15
Add MOABB Dataset Integration

milosobral:
## Summary

This PR introduces support for [MOABB (Mother of All BCI Benchmarks)](http://moabb.neurotechx.com/) datasets, adding two new EEG datasets with a reusable pipeline architecture.

### New Datasets

1. **PhysioNet Motor Imagery** (`schalk_wolpaw_physionet_2009`)
   - 109 subjects performing motor imagery tasks
   - Labels: left_hand, right_hand, hands, feet, rest

2. **Brain Invaders 2014a P300** (`korczowski_brain_invaders_2014a`)
   - 64 subjects performing visual P300 BCI task
   - Labels: Target, NonTarget

### New Infrastructure

- **`MOABBPipeline` base class** — Reusable pipeline for MOABB datasets handling:
  - Automatic manifest generation from dataset metadata
  - Data download via MOABB paradigm API
  - EEG extraction and trial interval creation
  - Stratified k-fold split generation

- **New task types** in taxonomy:
  - `MOTOR_IMAGERY`
  - `P300`

- **New split utilities**:
  - `generate_task_kfold_splits()` — Task-specific stratified cross-validation
  - `compute_subject_kfold_assignments()` — Deterministic subject-level k-fold assignments (hash-based)

## Dependencies

New pipeline dependencies:
- `moabb==1.4.3`
- `mne==1.11.0`
- `scikit-learn==1.8.0`

## Summary by CodeRabbit

* **New Features**
  * Added Motor Imagery and P300 task types.
  * Trial-level stratified k-folds with built-in validation splits and deterministic cross-subject fold assignments.
  * New dataset pipelines for BI2014a (P300) and PhysioNet Motor Imagery.
  * Reusable MOABB integration pipeline offering manifest generation, download/processing, trial extraction, splits, and end-to-end workflow.

* **Tests**
  * Expanded coverage for trial-level folds, per-task splits, subject-level assignments, and related error/warning cases.

### DIFF
--- a/brainsets/utils/moabb/pipeline.py
+++ b/brainsets/utils/moabb/pipeline.py
@@ -5,7 +5,6 @@ the brainsets pipeline framework. Subclasses define the dataset and paradigm
 classes, and implement paradigm-specific processing logic.
 """
 
-from abc import abstractmethod
 from typing import Dict, Any, Type, Optional
 from pathlib import Path
 from argparse import ArgumentParser
@@ -18,11 +17,6 @@ import mne
 
 from moabb.datasets.base import BaseDataset
 from moabb.paradigms.base import BaseParadigm
-from moabb.datasets.preprocessing import (
-    get_filter_pipeline,
-    get_resample_pipeline,
-    make_fixed_pipeline,
-)
 from temporaldata import Data, RegularTimeSeries, Interval, ArrayDict
 from brainsets.pipeline import BrainsetPipeline
 from brainsets import serialize_fn_map
@@ -36,26 +30,26 @@ from brainsets.taxonomy import Species, Task
 from brainsets.utils.split import generate_trial_folds
 
 
-_base_parser = ArgumentParser(add_help=False)
-_base_parser.add_argument(
+base_parser = ArgumentParser(add_help=False)
+base_parser.add_argument(
     "--redownload", action="store_true", help="Force redownload of raw data"
 )
-_base_parser.add_argument(
+base_parser.add_argument(
     "--reprocess", action="store_true", help="Force reprocessing of data"
 )
-_base_parser.add_argument(
+base_parser.add_argument(
     "--bandpass-low",
     type=float,
     default=None,
     help="Low cutoff frequency for bandpass filter in Hz (default: None, no filtering)",
 )
-_base_parser.add_argument(
+base_parser.add_argument(
     "--bandpass-high",
     type=float,
     default=None,
     help="High cutoff frequency for bandpass filter in Hz (default: None, no filtering)",
 )
-_base_parser.add_argument(
+base_parser.add_argument(
     "--resample",
     type=float,
     default=None,
@@ -68,6 +62,7 @@ class MOABBPipeline(BrainsetPipeline):
 
     Subclasses must define:
         - brainset_id: str
+        - brainset_description: BrainsetDescription
         - dataset_class: Type[BaseDataset]
         - paradigm_class: Type[BaseParadigm]
         - dataset_sign: str (e.g., "EEGBCI", "BRAININVADERS2014A")
@@ -80,9 +75,6 @@ class MOABBPipeline(BrainsetPipeline):
         - stratify_field: str (e.g., "movements", "targets")
         - label_map: Dict[str, int] (mapping from label strings to integer IDs)
 
-    Subclasses must implement:
-        - get_brainset_description(): Return dataset-specific BrainsetDescription
-
     The base class handles:
         - Manifest generation from dataset metadata
         - Data download via MOABB paradigm.get_data()
@@ -94,6 +86,7 @@ class MOABBPipeline(BrainsetPipeline):
     dataset_class: Type[BaseDataset]
     paradigm_class: Type[BaseParadigm]
     dataset_sign: str
+    brainset_description: BrainsetDescription
     dataset_kwargs: Dict[str, Any] = {}
     paradigm_kwargs: Dict[str, Any] = {}
 
@@ -213,13 +206,18 @@ class MOABBPipeline(BrainsetPipeline):
             )
 
     def download(self, manifest_item) -> Dict[str, Any]:
-        """Download and extract data using MOABB paradigm.
+        """Download and extract data using MOABB dataset.
+
+        Uses dataset._get_single_subject_data() instead of paradigm.get_data()
+        to avoid MOABB's SetRawAnnotations transformer, which assigns a fixed
+        duration to all annotations and silently drops the last event in each
+        run when that duration exceeds the recording length.
 
         This method:
         1. Sets up MNE download directory
         2. Validates bandpass parameters against Nyquist frequency
-        3. If no filtering requested: modifies paradigm to skip filtering step
-        4. Calls paradigm.get_data(return_raws=True) to get Raw objects
+        3. Loads Raw objects directly via dataset._get_single_subject_data()
+        4. Applies bandpass filtering and/or resampling if requested
         5. Filters results to the specific session from manifest_item
 
         Parameters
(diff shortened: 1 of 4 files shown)


brainsets/utils/split.py
```diff
@@ -325,3 +327,139 @@ def generate_stratified_folds(
             folds.append(fold_data)
 
     return folds
+
+
+def generate_task_kfold_splits(
```
estefanysuarez: It is not very clear what this and the next function (compute_subject_kfold_assignments) do.
Does this function create splits within sessions for a given tasks or across subjects?

I think these two functions would benefit from having an example in the docstrings. Also, if they do similar stuff, would it be worth it to maintain the same naming, meaning, using "generate" instead of "compute" in the second function. I have no preference of one over the other but just keep consistency across functions.

brainsets_pipelines/schalk_wolpaw_physionet_2009/pipeline.py
```diff
@@ -0,0 +1,113 @@
...
+    label_map = {
+        "left_hand": 0,
+        "right_hand": 1,
+        "hands": 2,
+        "feet": 3,
+        "rest": 4,
+    }
+
+    TASK_CONFIGS = {
+        "MotorImagery": ["left_hand", "right_hand", "hands", "feet", "rest"],
```
estefanysuarez: I have seen in several papers that when they use PhysionetMI for downstream tasks, they usually report a 4-class MI task, which I'm imagining is the "MotorImagery" here but without the label "rest" . Maybe we should have something like MotorImagery-4class and MotorImagery-5class, that exclude and include "rest" respectively?

brainsets/moabb_pipeline.py
```diff
@@ -0,0 +1,479 @@
...
+            recording_date = datetime.datetime(
+                2026, 1, 15, tzinfo=datetime.timezone.utc
+            )
+
+        return DeviceDescription(
+            id=f"{subject_id}_{recording_date.strftime('%Y%m%d')}",
+        )
+
+    def _generate_splits(self, trials, subject_id: Optional[str] = None):
+        """Generate stratified folds for trials.
```
estefanysuarez: You mean across trials? Maybe it is a bit nitpicky but "for trials" sounds like your splitting the interval/trial itself

estefanysuarez (commented):
Hey Milo, 
I went through the PR and I left some comments. Two things I'm not sure about:

1. Having a class for MOABB utils givent hat the other utils only have functions, does not seem consistent. Is there any particular advantage of having a class in this case?

2. This version of MOABB does not return all the trials at least for physionet. Not sure what the best solution to this is, it might not even be a problem. We can discuss this. 

One last thing that I think might be important to maybe add to the docstrings in the pipeline.py script is to specify which preprocessing steps the original data has gone through. I think by default MOABB applies a bandpass filter between 8-32Hz, not sure if we can get rid of that; it also cuts the trials to 3secs and sometimes it also applies referencing (not sure when it does and when it does not). 

The problem of using MOABB API's is that we are creating like a wrapper of a wrapper and we are not sure what they are doing with the data behind the scenes. One option that I considered at some point was to have the MOABB utils to get these datasets as MOABB if people want to compare to MOABB baselines, but maybe there should be a separate brainsets for Physionet that get the data from the original source.

estefanysuarez (commented):
Hey Milo!
Thanks for addressing the comments. This is what I noticed:
- Trials are no longer 3secs, which is good cause they are not cropped anymore
- I plot the PSD of the signal and it seems the bandpass filtering is effectively being removed
- My major comment would be in the timestamps of the trials (see details in `brainsets/utils/moabb/pipeline.py`). Maybe this is just a problem with Physionet and the fact that their API only returns 174 out of 180 trials (not sure why?). I didn't try with others, but I believe this should be fixed before merging. One thing I can tell you though, we were using this code to obtain the trials and I believe this was wrking correctly. Notice that the big difference here is in the definition of **epochs** and how single subject data is being obtained:

```
def extract_eeg_data(recording_data) -> RegularTimeSeries:
    sfreq = recording_data["0"].info["sfreq"]

    num_epochs = len(recording_data)
    epoch_data = [recording_data[str(i)].get_data().T for i in range(num_epochs)]
    eeg_signals = np.concatenate(epoch_data)

    # make regular time series
    eeg = RegularTimeSeries(
        signal=eeg_signals,
        sampling_rate=sfreq,
        domain=Interval(
            start=np.array([0.0]),
            end=np.array([(len(eeg_signals) - 1) / sfreq]),
        ),
    )

    units = ArrayDict(
        id=np.array(recording_data["0"].ch_names, dtype="U"),
        types=np.array(recording_data["0"].get_channel_types(), dtype="U"),
    )

    epoch_start = [0.0]
    epoch_end = []
    for m in epoch_data:
        if len(epoch_end) > 0:
            epoch_start.append(epoch_end[-1])
        epoch_end.append(epoch_start[-1] + len(m) / sfreq)

    epochs = Interval(
        start=np.array(epoch_start),
        end=np.array(epoch_end),
    )

    return eeg, units, epochs


def extract_motor_imagery_trials(recording_data, epochs) -> Interval:
    start_times = []
    end_times = []
    movements = []

    for i in range(len(epochs)):
        epoch_start = epochs.start[i]

        for annotation in recording_data[str(i)].annotations:
            start_time = annotation["onset"] + epoch_start
            end_time = start_time + annotation["duration"]

            if len(end_times) > 0 and start_time < end_times[-1]:
                # the previous trial goes into the next one
                # this happens because of numerical precision issues
                assert (
                    end_times[-1] - start_time < 0.1
                ), f"found overlap between trials: start time of trial i: {start_time}, end time of trial i-1: {end_times[-1]}"

                # we can clip the end time of the last trial
                end_times[-1] = start_time

            start_times.append(start_time)
            end_times.append(end_time)
            movements.append(annotation["description"])

    trials = Interval(
        start=np.array(start_times),
        end=np.array(end_times),
        timestamps=(np.array(start_times) + np.array(end_times)) / 2,
        movements=np.array(movements),
        movement_ids=np.array([MOVEMENT_ID_MAP[m] for m in movements]),
        timekeys=["start", "end", "timestamps"],
    )

    if not trials.is_disjoint():
        raise ValueError("Found overlapping trials")

    return trials
```

Where data is obtained in the following way:

```
    dataset = PhysionetMI(imagined=True, executed=False)
    
    # load data
    dataset.get_data(subjects=[args.subject])

    recording_data = dataset._get_single_subject_data(args.subject)["0"]
```

brainsets_pipelines/korczowski_brain_invaders_2014a/pipeline.py
```diff
@@ -0,0 +1,93 @@
...
+
+from brainsets.descriptions import BrainsetDescription
+from brainsets.taxonomy import Task
+from brainsets.utils.moabb.pipeline import MOABBPipeline, _base_parser
+from brainsets.utils.split import generate_subject_kfold_assignment
+
+
+logging.basicConfig(level=logging.INFO)
+
+parser = ArgumentParser(parents=[_base_parser])
```
vinamarora8: `_base_parser` should not have an underscore at the start, since it's intended to be a public object.

brainsets/utils/moabb/pipeline.py
```diff
@@ -0,0 +1,648 @@
...
+        - paradigm_class: Type[BaseParadigm]
+        - dataset_sign: str (e.g., "EEGBCI", "BRAININVADERS2014A")
+        - dataset_kwargs: Dict[str, Any] (optional, defaults to {})
+        - paradigm_kwargs: Dict[str, Any] (optional, defaults to {})
+        - task: Task (e.g., Task.MOTOR_IMAGERY, Task.P300)
+        - trial_key: str (e.g., "motor_imagery_trials", "p300_trials")
+        - label_field: str (e.g., "movements", "targets")
+        - id_field: str (e.g., "movement_ids", "target_ids")
+        - stratify_field: str (e.g., "movements", "targets")
+        - label_map: Dict[str, int] (mapping from label strings to integer IDs)
```
vinamarora8: Could this be validated in the pipeline's `__init__`?

brainsets/utils/moabb/pipeline.py
```diff
@@ -0,0 +1,648 @@
...
+
+    Subclasses must implement:
+        - get_brainset_description(): Return dataset-specific BrainsetDescription
+
```
vinamarora8: Does BrainsetDescription need to be programmatically generated? If not, then it could be made into a constant class attribute like the other attributes.

brainsets/utils/moabb/pipeline.py
```diff
@@ -0,0 +1,637 @@
...
+                for fmin, fmax in paradigm.filters:
+                    combined = make_fixed_pipeline(
+                        get_filter_pipeline(fmin, fmax),
+                        get_resample_pipeline(resample_rate),
+                    )
+                    pipelines.append(combined)
+                return pipelines
+
+            paradigm._get_raw_pipelines = make_filter_resample_pipelines
+
```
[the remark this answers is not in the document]
bruAristimunha: it is okay use private at the moment

brainsets/utils/moabb/pipeline.py
```diff
@@ -0,0 +1,479 @@
...
+        Returns
+        -------
+        list[str]
+            List of channel type strings (e.g., "EEG", "EOG", "EMG")
+        """
+        if len(epochs) > 0:
+            return epochs[0].get_channel_types()
+
+        # MNE channel kind constants (mne.io.constants.FIFF)
```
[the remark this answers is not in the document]
bruAristimunha: These values come from mne itself, which is how mne organizes the channel types. More details can be find within the FIFF specification.

brainsets/utils/moabb/pipeline.py
```diff
@@ -0,0 +1,648 @@
...
+    @abstractmethod
+    def get_brainset_description(self) -> BrainsetDescription:
+        """Return dataset-specific BrainsetDescription.
+
+        Returns
+        -------
+        BrainsetDescription
+            Description object with dataset metadata
+        """
+        ...
```
bruAristimunha: You may be able to get this from moabb itself, but only when moabb is released.

brainsets/utils/moabb/pipeline.py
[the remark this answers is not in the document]
estefanysuarez: Awesome, thanks @bruAristimunha! I will take a look at that.

bruAristimunha (approved):
LGTM, very small comments!

PR #75, opened 2026-01-19
Open Neuro Base Class + Example

milosobral:
### Add OpenNeuro EEG Base Pipeline

This PR introduces a reusable `OpenNeuroEEGPipeline` base class that handles common boilerplate for processing BIDS-compliant EEG datasets from OpenNeuro's S3 bucket.

#### What's Added

**Core Components (`brainsets/utils/openneuro/`)**
- `dataset.py` — S3-based file listing, BIDS filename parsing to auto-discover EEG recordings, and download utilities
- `data_extraction.py` — Extract metadata (subject, session, device descriptions) and signals from MNE Raw objects into brainsets data structures
- `pipeline.py` — `OpenNeuroEEGPipeline` base class providing default `get_manifest()`, `download()`, and `process()` implementations

**Supporting Utilities**
- `bids_utils.py` — BIDS EEG filename parser
- `mne_utils.py` — Modality-to-MNE channel type mapping
- Enhanced `s3_utils.py` — Generic S3 download functions with caching

**Example Pipeline: `klinzing_sleep_ds005555_2024`**

Demonstrates how minimal a new pipeline can be (~50 lines):

```python
class Pipeline(OpenNeuroEEGPipeline):
    brainset_id = "klinzing_sleep_ds005555_2024"
    dataset_id = "ds005555"
    
    ELECTRODE_RENAME = {"PSG_F3": "F3", "PSG_F4": "F4", ...}
    MODALITY_CHANNELS = {"EEG": ["F3", "F4", ...], "EOG": ["EOG"], ...}
```

#### Key Features
- **Auto-discovery**: Automatically finds all EEG recordings in a dataset by querying OpenNeuro S3 and parsing BIDS filenames
- **Smart caching**: Skips re-downloading/re-processing with `--redownload` and `--reprocess` flags
- **Channel mapping**: Declarative electrode renaming and modality assignment via class attributes
- **Extensible**: Override `process()` for dataset-specific logic while keeping the common workflow

## Summary by CodeRabbit

* **New Features**
  * Added three OpenNeuro dataset pipelines and dataset classes (sleep, visual-naming, HBN-R1), plus a consolidated OpenNeuro utilities interface for dataset discovery, download, and BIDS parsing.
  * Broadened age normalization to accept more numeric types.

* **Tests**
  * Added extensive tests for OpenNeuro utilities, pipelines, BIDS parsing, participants handling, and S3 client behavior.

* **Chores**
  * Exposed explicit S3 client cache control, updated test dependencies and pytest config, and updated changelog.


brainsets/utils/openneuro/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+    "extract_subject_description",
+    "extract_session_description",
+    "extract_device_description",
+    "extract_meas_date",
+    "extract_signal",
+    "extract_channels",
+    "generate_train_valid_splits_one_epoch",
+    # pipeline.py
+    "OpenNeuroEEGPipeline",
+]
```
vinamarora8: Just curious as to why use an `__all__` here, since basically you are exporting everything in this module. I've seen people do that, so am just curious behind the thought.

brainsets/utils/openneuro/data_extraction.py
```diff
@@ -0,0 +1,230 @@
...
+    Returns:
+        BrainsetDescription object
+    """
+    return BrainsetDescription(
+        id=dataset_id,
+        origin_version=origin_version,
+        derived_version=derived_version,
+        source=source,
+        description=description,
+    )
```
vinamarora8: Seems quite redundant

brainsets/utils/openneuro/data_extraction.py
```diff
@@ -0,0 +1,230 @@
...
+    """
+    return ArrayDict(
+        id=np.array(recording_data.ch_names, dtype="U"),
+        types=np.array(recording_data.get_channel_types(), dtype="U"),
+    )
+
+
+def generate_train_valid_splits_one_epoch(
+    epoch: Interval, split_ratios: list[float] = None
+) -> tuple[Interval, Interval]:
```
vinamarora8: Since it's only splitting into two; maybe a better API would be to ask for `train_ratio` directly instead of `split_ratios`. Since they must sum to 1.0, only one float is sufficient to specify.

brainsets/utils/bids_utils.py
```diff
@@ -0,0 +1,45 @@
...
+    r"^(?P<subject>sub-[^_]+)"
+    r"(?:_(?P<session>ses-[^_]+))?"
+    r"_task-(?P<task>[^_]+)"
+    r"(?:_acq-(?P<acq>[^_]+))?"
+    r"(?:_run-(?P<run>[^_]+))?"
+    r"_eeg"
+    r"\.(?P<ext>\w+)$"
+)
+
+EEG_EXTENSIONS = {".edf", ".vhdr", ".set", ".bdf"}
```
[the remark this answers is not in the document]
oumgharbi: also there are some datasets with other files such as `.mat` files (e.g. [ds001512](https://openneuro.org/datasets/ds002739/versions/1.0.0)). Do you think we should add them too ?

brainsets/utils/openneuro/pipeline.py
```diff
@@ -0,0 +1,454 @@
...
+                raise ValueError(f"Could not find .vhdr header file for {eeg_file}")
+            eeg_file = vhdr_file
+
+        try:
+            return mne.io.read_raw(eeg_file, preload=True, verbose=False)
+        except Exception as e:
+            raise RuntimeError(f"Failed to load EEG file {eeg_file}: {e}") from e
+
+    def _process_common(self, download_output: dict) -> tuple[Data, Path]:
+        """Process EEG files and create a Data object.
```
vinamarora8: Should this be a private method? I think if a user overrides `process`, they'll end up using this method in it. So, I guess it should be public?

brainsets_pipelines/klinzing_sleep_ds005555_2024/pipeline.py
```diff
@@ -0,0 +1,76 @@
...
+}
+
+PSG_MODALITY_CHANNELS = {
+    "EEG": ["F3", "F4", "C3", "C4", "O1", "O2"],
+    "EOG": ["EOG", "EOGL", "EOGR"],
+    "EMG": ["EMG"],
+    "RESPIRATORY": ["THER", "THOR", "ABD", "CAN"],
+    "PPG": ["PULSE", "BEAT", "SPO2"],
+}
+
```
oumgharbi: Not required but would be nice to have: the information on participants (age, sex, etc.) are usually in the `participants.tsv` and rarely found on the raw data files. We can eventually add this with the agent-based operations.

oumgharbi (commented):
Hey Milo, great work. It is more structured now and certainly independent from the agent's output. I added some minor comments.
I still have a concern about the channel modality. Although datasets are required to provide the modality of each channel, I think we should not take it for granted and it would be best to add a check of the raw signals on our side. This stays optional of course.
 I think this pipeline will work for most of OpenNeuro datasets. There will be some edge cases that will require some tweeks.

oumgharbi:
@milosobral would it be possible to open a separate PR to add the `brainsets/utils/s3_utils.py` so we can reuse it asap for other pipelines?
It would also be useful to get the `data_extraction.py` out of `brainsets/utils/openneuro/` as it has no real dependency on openneuro pipelines. maybe have it as is or as `mne_utils.py`.
What do you think?

brainsets/utils/bids_utils.py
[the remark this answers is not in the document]
milosobral: Honestly I think for now we should ignore non-BIDS compliant datasets

brainsets/utils/openneuro/pipeline.py
```diff
@@ -0,0 +1,606 @@
...
+            recording_id: Recording identifier (not used; metadata comes from raw)
+            data_dir: Data directory (not used; metadata comes from raw)
+
+        Returns:
+            ArrayDict with channel id, type, status, and coordinate fields
+        """
+        channel_ids = np.array(raw.ch_names, dtype="U")
+        channel_count = len(channel_ids)
+
+        channel_types = np.array(raw.get_channel_types(), dtype="U")
```
oumgharbi: We will be running into the same problem with EEG channels, where all channels will have `eeg` types, even when they're actually not (ECoG, sEEG, etc.)
Maybe a workaround would be to extract such metadata from the side files like `*_channels.tsv`, which are usually present.

brainsets/utils/openneuro/pipeline.py
```diff
@@ -0,0 +1,606 @@
...
+        for idx, ch_name in enumerate(channel_ids):
+            if ch_name in bad_channels:
+                status[idx] = "bad"
+
+        x_coords = np.full(channel_count, np.nan)
+        y_coords = np.full(channel_count, np.nan)
+        z_coords = np.full(channel_count, np.nan)
+
+        try:
+            montage = raw.get_montage()
```
oumgharbi: Same as above. Many datasets don't embed this info in the raw data, but provide it elsewhere, usually in `*_electrodes.tsv`.

brainsets/utils/openneuro/data_extraction.py
oumgharbi: We can entirely remove the `extract_*_description` utils here, refractor the `read_bids_*` utils to the `utils/bids_utils.py` and completely remove this `data_extraction.py`.

brainsets/utils/bids_utils.py
```diff
@@ -0,0 +1,159 @@
...
+)
+
+# BIDS iEEG supported formats (BIDS v1.10.1):
+# - European Data Format (.edf): Single file per recording.
+# - BrainVision (.vhdr): Header file; requires .vmrk (markers) and .eeg (data) files.
+# - EEGLAB (.set): MATLAB format; optional .fdt file contains float data.
+# - Biosemi (.bdf): Single file per recording.
+# - NWB (.nwb): Neurodata Without Borders format for standardized neurophysiology storage.
+# Reference: https://bids-specification.readthedocs.io/en/stable/modality-specific-files/intracranial-electroencephalography.html
+IEEG_EXTENSIONS = {".edf", ".vhdr", ".set", ".bdf", ".nwb"}
```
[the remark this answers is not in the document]
milosobral: The reason why these extensions are not included is that they are not the extensions of the main files. vmrk for example is the file that will contain markers that comes alongside a vhdr file. These extensions are used to get the manifest, so I need to make sure that we only count each recording once, and the easiest way to do that is only looking for the main file extensions.

brainsets/utils/openneuro/pipeline.py
```diff
@@ -0,0 +1,609 @@
...
+
+class OpenNeuroPipeline(BrainsetPipeline, ABC):
+    """Abstract base class for OpenNeuro dataset pipelines.
+
+    This class extends BrainsetPipeline and provides common functionality for
+    processing datasets from OpenNeuro, supporting both EEG and iEEG modalities.
+
+    **Required class attributes:**
+        - dataset_id: OpenNeuro dataset identifier (e.g., "ds005555")
+        - brainset_id: Unique identifier for the brainset
```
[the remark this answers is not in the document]
milosobral: I think given that this is a brainsets wide thing and not an open neuro specific thing, it probably doesn't belong in this PR

oumgharbi (commented):
Thanks Milo for the updates. I like the idea of having the `OpenNeuroPipeline` base then having another base for each modality. 
I added some comments, and I think you already started adressing them 😅

brainsets/utils/openneuro/pipeline.py
milosobral: "Maybe a workaround would be to extract such metadata from the side files like *_channels.tsv, which are usually present."
I think this is what is happening here. MNE-bids is reading those sidecar files to get this information. I am making the assumption here that iEEG datasets will not have the same issues that EEG dataset and that they're metadata is higher quality and more reliable. This is based on the fact that they require some surgical intervention so the standards for data quality and integrity is probably higher.

brainsets/utils/openneuro/pipeline.py
```diff
@@ -0,0 +1,601 @@
...
+        self.update_status("Extracting Metadata")
+        source = f"https://openneuro.org/datasets/{self.dataset_id}"
+        dataset_description = (
+            self.description
+            if self.description
+            else f"OpenNeuro dataset {self.dataset_id}"
+        )
+
+        brainset_description = BrainsetDescription(
+            id=self.dataset_id,
```
estefanysuarez: Align with the 🐰, I believe this has to correspond to `brainset_id` instead of `dataset_id`. Maybe this is totally unrelated, but @milosobral remember we were having issues at some point because of some mismatch between the id in `BrainsetDescription.id` and, I believe but don't quote me on this, the name of the processed folder in which .h5 files are stored?

brainsets/utils/openneuro/pipeline.py
estefanysuarez: I'm not entirely sure about higher standards in terms of data formatting (for sure yes in terms of data quality!). However, I think this is precisely one of those things that we should check when creating a brainset. Which reminds ☝🏼 we should come up with a minimal checklist/rubric to use when going through open neuro datasets for creating new brainsets.

brainsets/utils/openneuro/pipeline.py
```diff
@@ -0,0 +1,601 @@
...
+        for idx, ch_name in enumerate(channel_ids):
+            if ch_name in bad_channels:
+                status[idx] = "bad"
+
+        x_coords = np.full(channel_count, np.nan)
+        y_coords = np.full(channel_count, np.nan)
+        z_coords = np.full(channel_count, np.nan)
+
+        try:
+            montage = raw.get_montage()
```
estefanysuarez: Q: Are there MNE montages for iEEG data? I did a quick search and I found that MNE has this special funtion for iEEG data: [mne.preprocessing.ieeg.make_montage_volume()](https://mne.tools/stable/generated/mne.preprocessing.ieeg.make_montage_volume.html)

estefanysuarez (commented):
Hey @milosobral @oumgharbi !

Congratulations on all this amazing work. I think this is great and hopefully we will be able to integrate more brainsets soon 🚀

I just have two suggestions:
- You will see the comments in `brainsets/utils/openneuro/pipeline.py` but I believe being consistent with the versioning system of Open Neuro is very important for reproducibility reasons. 

- One thing that I believe we should incorporate, ideally not only for OpenNeuro, but for all brainsets, and that could be super helpful for researchers, OpenNeuro users in general, as well as for ON contributors, is if we could create a `.bib` file that contains the citations for Open Neuro plus for the downloaded datasets. It does not have to be in this PR but I think we should think of something like this. Happy to discuss about it! 

- I'm writing this here more as a note to remind myself that we should come up with a sort of rubric or minimal checklist of the things that we need to verify when creating brainsets for OpenNeuro. 

Great job guys!

brainsets/utils/openneuro/pipeline.py
```diff
@@ -0,0 +1,601 @@
...
+        source = f"https://openneuro.org/datasets/{self.dataset_id}"
+        dataset_description = (
+            self.description
+            if self.description
+            else f"OpenNeuro dataset {self.dataset_id}"
+        )
+
+        brainset_description = BrainsetDescription(
+            id=self.dataset_id,
+            origin_version=f"openneuro/{self.brainset_id}",
```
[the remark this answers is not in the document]
estefanysuarez: Based on our conversation, my preference would be:
- Always allow users to get the latest version just because in theory if there is an update in the data it is supposed to be better
- Throw a warning to the user making them aware that they are using the latest version and provide the version tag. 
- Because changes in datasets are unpredictable, there could be a second warning regarding the version tag used when the brainset was built and also adding a suggestion to the user to verify what changed in the dataset. This may or not affect the current brainsets pipeline.
- In an ideal world, we should somehow allow users to download a specific version tag (see `DataLad`)

oumgharbi (commented):
Thank you for the updates @milosobral 
The new dataset classes may be a bit redundant. What do you think about creating an  `OpenNeuroDataset` API, defining a reusable base class that inherits from `Dataset`, and implements the constructor and sampling intervals logic. Dataset-specific details can be set as attributes in thin subclasses. 
Let me know what you think.

milosobral:
Adressed in #128

PR #76, opened 2026-01-19
Adding the Child Mind Institute MIPDB to brainsets

oumgharbi:
Adding this EEG dataset intitled [Multimodal Resource for Studying Information Processing in the Developing Brain (MIPDB)
](https://fcon_1000.projects.nitrc.org/indi/cmi_eeg/index.html) from 126 participants mostly children and young adults, recorded at the Child Mind Institute.


oumgharbi:
The MIPDB dataset contains 3 modalities: EEG, Eyetracking and behavioral data. The current pipeline gets and prepares EEG data. I will be adding the other modalities soon.

brainsets_pipelines/kelly_CMI_developingbrain_2016/pipeline.py
```diff
@@ -0,0 +1,401 @@
...
+    86: ("Naturalistic Viewing Paradigm Video 6", Task.NATURALISTIC_VIEWING),
+}
+
+parser = ArgumentParser()
+parser.add_argument("--redownload", action="store_true")
+parser.add_argument("--reprocess", action="store_true")
+
+
+class Pipeline(BrainsetPipeline):
+    brainset_id = "kelly_CMI_developingbrain_2016"
```
vinamarora8: Maybe we should call this brainset `cmi_mipdb_2016`? Since MIPDB seems to be its colloquial name. I know this doesn't follow the convention written in the docs, but we did something similar with `allen_visual_coding_ophys_2016`.

vinamarora8 (commented):
Left some early discussion comments

Also, can you please add one-session processing of this dataset to the `.github/workflows/pipeline_testing.yml`?

brainsets_pipelines/kelly_CMI_developingbrain_2016/pipeline.py
```diff
@@ -0,0 +1,401 @@
...
+    parser = parser
+
+    def _download_subject_metadata(self) -> Path:
+        """Download the MIPDB public metadata CSV if not already cached locally.
+
+        Returns the local path to the downloaded CSV file.
+        """
+        # the metadata file is not hosted in the same s3 bucket
+        local_path = self.raw_dir / "MIPDB_PublicFile.csv"
+        metadata_url = "https://fcon_1000.projects.nitrc.org/indi/cmi_eeg/_static/MIPDB_PublicFile.csv"
```
milosobral: I don't love constants like this being hidden inside of a function. I think you could define it as a constant at the beggining of the file with uppercas naming

brainsets_pipelines/kelly_CMI_developingbrain_2016/pipeline.py
```diff
@@ -0,0 +1,401 @@
...
+            id=subject_id,
+            species="HUMAN",
+            age=subject_info["age"],
+            sex=subject_info["sex"],
+        )
+
+        self.update_status("Extracting EEG signal and channels")
+        eeg_signal = extract_eeg_signal(raw)
+        channels = extract_channels(raw)
+        self._add_channel_locations(channels)
```
milosobral: Nitpick: I tend to believe that in instances like this, doing things in place might add a bit of confusion, especially since it is not really common in brainsets for utility functions like this. This seems cleaner to me: `channels.locations = self._get_channel_locations(channels)`
This is very much a nitpick and a matter of opinion so feel free to ignore and resolve this comment if you don't agree.

brainsets_pipelines/kelly_CMI_developingbrain_2016/pipeline.py
```diff
@@ -0,0 +1,401 @@
...
+
+        self.update_status("PROCESSING")
+
+        recording_id = raw_path.stem
+
+        output_path = self.processed_dir / "EEG" / f"{recording_id}.h5"
+        if output_path.exists() and not self.args.reprocess:
+            self.update_status(f"Skipping processing, file exists: {output_path}")
+            return
+
```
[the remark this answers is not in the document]
oumgharbi: I will actually remove this, because I will add behavioral data along with EEG, in the same h5 file!

oumgharbi:
Hey all, 
I updated this brainset pipeline, mainly to create splits according to our agreed splitting strategy.
The create_splits function now produces two distinct categories of splits:

1. Behavior-agnostic splits (always present, no folds, train/valid only)
- Intrasession: causal split.
- Intersubject: deterministic subject-level train/valid assignment.
- Intersession: deterministic session-level train/valid assignment.

2. Behavior-relevant splits (only when paradigm annotations exist, 3-fold, train/valid/test)

- Intrasession: stratified k-fold by task, ensuring balanced paradigm representation across folds.
- Intersubject: per-fold subject-level train/valid/test assignment.
- Intersession: per-fold session-level train/valid/test assignment.

All intervals are subdivided into 30s epochs before splitting. (we can discuss if this makes sense)

The `CmiMIPDB2016` class is also updated accordingly, with a new `split_category` parameter (`"behavior-agnostic"`/`"behavior-relevant" `)

Below is an example of an EEG recording with eye-tracking data (samples only plotted), comprising 3 paradigms (Contrast change detection, and two naturalistic viewing paradigms).

Let me know what you think!


A00053375006_paradigms

brainsets_pipelines/cmi_mipdb_2016/constants.py
```diff
@@ -0,0 +1,78 @@
+from brainsets.taxonomy import Task
+
```
milosobral, line 2: If there is some documentation somewhere explaining those, it would be cool to link it here.

brainsets_pipelines/cmi_mipdb_2016/constants.py
```diff
@@ -0,0 +1,78 @@
...
+    "R Validity": "r_validity",
+    "Pupil Confidence": "pupil_confidence",
+    "L Plane": "l_plane",
+    "R Plane": "r_plane",
+    # "H POS X [mm]": "h_pos_x",
+    # "H POS Y [mm]": "h_pos_y",
+    # "H POS Z [mm]": "h_pos_z",
+    # "H ROT X [°]": "h_rot_x",
+    # "H ROT Y [°]": "h_rot_y",
+    # "H ROT Z [°]": "h_rot_z",
```
milosobral, line 65: Any reason why those are commented out?

brainsets_pipelines/cmi_mipdb_2016/paradigm.py
```diff
@@ -0,0 +1,224 @@
...
+        return int(float(str(annotations.description[i]).strip()))
+    except (ValueError, TypeError):
+        return None
+
+
+def _paradigm_segment_end(
+    start_idx: int,
+    paradigm_code: int,
+    annotations: Interval,
+    paradigm_start_indices: list[int],
```
milosobral, line 25: Very nitpick but for the sake of keeping the interface of each function consistent, could you keep `annotations: Interval` as the first argument like you did in the other helper functions?

brainsets_pipelines/cmi_mipdb_2016/paradigm.py
```diff
@@ -0,0 +1,224 @@
...
+        )
+
+    # Last annotation before next paradigm start, or last in recording
+    try:
+        pos = paradigm_start_indices.index(start_idx)
+        next_idx = (
+            paradigm_start_indices[pos + 1]
+            if pos + 1 < len(paradigm_start_indices)
+            else None
+        )
```
milosobral, line 58: If it is the line above this (`pos = paradigm_start_indices.index(start_idx)`) that throws the ValueError, maybe this could be outside ot the try except block

brainsets_pipelines/cmi_mipdb_2016/paradigm.py
```diff
@@ -0,0 +1,224 @@
...
+def _extract_segment_events(
+    annotations: Interval,
+    start_s: float,
+    end_s: float,
+) -> tuple[list[int], list[float], list[str]]:
+    """Return all numeric annotation events within ``[start_s, end_s]``."""
+    mask = (annotations.start >= start_s) & (annotations.start <= end_s)
+    codes: list[int] = []
+    onsets: list[float] = []
+    descriptions: list[str] = []
```
milosobral, line 75: I think you could use `Interval.slice()` here

brainsets_pipelines/cmi_mipdb_2016/paradigm.py
```diff
@@ -0,0 +1,224 @@
...
+        if code is not None:
+            codes.append(code)
+            onsets.append(float(annotations.start[idx]))
+            descriptions.append(str(annotations.description[idx]))
+    return codes, onsets, descriptions
+
+
+def get_paradigm_interval_for_code(
+    paradigm_code: int,
+    annotations: Interval,
```
milosobral, line 87: Same nitpick here, could make annotations first argument to stay consistent

brainsets_pipelines/cmi_mipdb_2016/paradigm.py
```diff
@@ -0,0 +1,224 @@
...
+    Args:
+        paradigm_code: Paradigm start code (must be in PARADIGM_MAP).
+        annotations: Raw annotations from the recording (start, end, description).
+
+    Returns:
+        Interval with one segment per occurrence of this paradigm (start, end,
+        description, code, task). Empty Interval if the paradigm does not appear.
+    """
+    if paradigm_code not in PARADIGM_MAP or len(annotations) == 0:
+        return Interval(start=np.array([]), end=np.array([]))
```
milosobral, line 100: Is there a reason why we would want to allow a non-existent paradigm code to silently fail? Maybe throwing an error when the Paradigm code is not valid is reasonable to avoid some silent errors.

brainsets_pipelines/cmi_mipdb_2016/paradigm.py
```diff
@@ -0,0 +1,224 @@
...
+        start_code=np.array(codes, dtype=np.int64),
+        task=np.array(tasks, dtype="U"),
+    )
+
+
+def get_paradigm_events(
+    annotations: Interval,
+    paradigm_intervals: Interval,
+) -> Interval:
+    """Build a flat event table linked to paradigm rows via ``paradigm_idx``."""
```
milosobral, line 185: This docstring is missing some details about what the arguments should be and what is returned

brainsets_pipelines/cmi_mipdb_2016/eyetracking.py
```diff
@@ -0,0 +1,695 @@
...
+from pathlib import Path
+
+import numpy as np
+import pandas as pd
+from temporaldata import Data, Interval, IrregularTimeSeries
+
+from constants import PARADIGM_MAP, SAMPLES_COLUMNS
+
+
+def parse_comma_separated_numbers(s: str) -> list[int | float]:
```
milosobral, line 30: This should probably be a private method

brainsets_pipelines/cmi_mipdb_2016/eyetracking.py
```diff
@@ -0,0 +1,695 @@
...
+                paradigm_intervals.event_codes[paradigm_idx]
+            )
+            event_onsets = parse_comma_separated_numbers(
+                paradigm_intervals.event_onsets[paradigm_idx]
+            )
+            grouped[paradigm_idx] = [
+                (float(onset), int(code))
+                for onset, code in zip(event_onsets, event_codes)
+            ]
+        return grouped
```
milosobral, line 93: Maybe I'm missing something but any reason why you still support some "legacy" structure? Couldn't you simply make sure to always use the new one or is the old one still necessary because of the dataset?

brainsets_pipelines/cmi_mipdb_2016/eyetracking.py
```diff
@@ -0,0 +1,695 @@
...
+                continue
+            grouped[paradigm_idx].append(
+                (
+                    float(paradigm_events.start[idx]),
+                    int(paradigm_events.code[idx]),
+                )
+            )
+        for events in grouped:
+            events.sort(key=lambda item: item[0])
+        return grouped
```
milosobral, line 77: Maybe here you could use some of the interval methods to make this a bit easier. Using `select_by_mask` would maybe make the code a bit more readable

brainsets_pipelines/cmi_mipdb_2016/eyetracking.py
```diff
@@ -0,0 +1,695 @@
...
+                    try:
+                        ts = int(parts[3])
+                    except ValueError:
+                        continue
+                    user_events.append((ts, parts[4].strip()))
+    return user_events
+
+
+def _identify_segments(user_events: list[tuple[int, str]]) -> list[dict]:
+    """Identify paradigm segments bounded by ``start_eye_recording`` markers.
```
milosobral, line 120: I think the docstring should specify that `user_events` is supposed to be read from an eye tracking file. It makes sense given that this function is in eyetracking.py but this wouldn't work with a user_event list from the EEG files

brainsets_pipelines/cmi_mipdb_2016/eyetracking.py
```diff
@@ -0,0 +1,695 @@
...
+    index: list[dict] = []
+    for base_id in sorted(bases.keys()):
+        entry = bases[base_id]
+        samples_path = entry["samples_path"]
+        events_path = entry["events_path"]
+
+        if events_path is not None and events_path.exists():
+            user_events = _parse_user_events_from_file(events_path)
+        elif samples_path is not None and samples_path.exists():
+            user_events = _parse_user_events_from_file(samples_path)
```
milosobral, line 219: could you maybe add a short comment explaining why events take precedence of samples here?

brainsets_pipelines/cmi_mipdb_2016/eyetracking.py
```diff
@@ -0,0 +1,695 @@
...
+    while r < n and c < m:
+        if seq_a[r] == seq_b[c]:
+            pairs.append((r, c))
+            r += 1
+            c += 1
+        elif dp[r + 1][c] >= dp[r][c + 1]:
+            r += 1
+        else:
+            c += 1
+    return pairs
```
milosobral, line 266: Not 100% sure but could we maybe use something like this https://docs.python.org/3/library/difflib.html instead of having this complex function here? 
I trust that it works but it is not the easiest thing to understand and means a new relatively complex thing to maintain.

milosobral (commented):
Great work, especially considering the  complexity of the Eye tracking stuff and the mess with multiple files with non matching timelines...
I haven't had time to go through everything or to test the pipeline and the data yet but here are some initial comments from reading the code. 

(Note that I still need to go through the second half of eye_tracking.py, pipeline.py, and the dataset stuff)

brainsets/datasets/CmiMIPDB2016.py
```diff
@@ -0,0 +1,174 @@
...
+from temporaldata import Data
+
+from torch_brain.dataset import Dataset
+
+from ._utils import get_processed_dir
+
+SplitCategory = Literal["behavior_agnostic", "behavior_relevant"]
+VALID_SPLIT_CATEGORIES = get_args(SplitCategory)
+
+FoldType = Literal["intrasession", "intersubject", "intersession"]
```
estefanysuarez, line 13: Even though we decided to do not impose any standardization on these kind of things, I suggest to rename `FoldType` and its name in the signature of the constructor function (`fold_type`) to `SplitType` and `split_type`, respectively? Just to make it more consistent with other Datasets.

brainsets/datasets/CmiMIPDB2016.py
```diff
@@ -0,0 +1,174 @@
...
+VALID_FOLD_TYPES = get_args(FoldType)
```
estefanysuarez, line 14: Same here, replace `VALID_FOLD_TYPES` with `VALID_SPLIT_TYPES`

brainsets/datasets/CmiMIPDB2016.py
```diff
@@ -0,0 +1,174 @@
...
+
+N_FOLDS = 3
+
+
+class CmiMIPDB2016(Dataset):
+    """CMI Multimodal Resource for Studying Information Processing in the
```
estefanysuarez, line 20: In the docstring I would explicitly spell Child Mind Institute and then use the abbreviation:
`"""Child Mind Institute (CMI) Multimodal Resource for Studying ..."""`

brainsets/datasets/CmiMIPDB2016.py
```diff
@@ -0,0 +1,174 @@
...
+
+    Args:
+        root: Root directory for the dataset.
+        recording_ids: List of recording IDs to load.
+        transform: Data transformation to apply.
+        uniquify_channel_ids: Whether to prefix channel IDs with session ID
+            to ensure uniqueness. Defaults to True.
+        split_category: Top-level split category. Must be one of:
+
+            - ``"behavior_agnostic"``: No folds, train/valid only. Always available for all recordings.
```
estefanysuarez, line 37: Maybe add a sentence that explicitly states the main purpose of this split category. Something along the lines: 

`- ``"behavior_agnostic"``: No folds; includes only train and validation splits. Available across all recordings, and primarily intended for foundation model pretraining.`

brainsets/datasets/CmiMIPDB2016.py
```diff
@@ -0,0 +1,174 @@
...
+            - ``"behavior_relevant"``: K-fold with train/valid/test. Only available for recordings that have paradigm annotations.
+
+            Defaults to ``"behavior_agnostic"``.
+        fold_type: The splitting strategy. Must be one of:
```
estefanysuarez, line 41: Following my previous comments, replace `fold_type` for `split_type`

brainsets/datasets/CmiMIPDB2016.py
```diff
@@ -0,0 +1,174 @@
...
+                raise ValueError(
+                    f"fold_number must be between 0 and {N_FOLDS - 1}, "
+                    f"got {fold_number}"
+                )
+        self.fold_number = fold_number
+
+    def get_sampling_intervals(
+        self,
+        split: Optional[Literal["train", "valid", "test"]] = None,
+    ):
```
estefanysuarez, line 106: Not sure what would be the best practice here since this is a method that exists already in the main `Dataset` parent class and is well documented, but adding some short docstrings that explain `Args` and `Returns` sections would be helpful

brainsets/datasets/CmiMIPDB2016.py
```diff
@@ -0,0 +1,174 @@
...
+            return self._get_default_intervals(split)
+        else:
+            if split not in ("train", "valid", "test"):
+                raise ValueError(
+                    f"Invalid split '{split}'. "
+                    f"Must be one of ['train', 'valid', 'test']."
+                )
+            return self._get_behavior_intervals(split)
+
+    def _get_default_intervals(self, split: str) -> dict:
```
estefanysuarez, line 125: To be consistent throughout, I would rename this method to `_get_behavior_agnostic_intervals()`. I would also add a brief docstrings.

brainsets/datasets/CmiMIPDB2016.py
```diff
@@ -0,0 +1,174 @@
...
+            if not hasattr(rec.splits, "behavior_relevant"):
+                continue
+            assignment = str(rec.get_nested_attribute(assignment_key))
+            if assignment == split:
+                result[rid] = rec.domain
+        return result
+
+    def get_recording_hook(self, data: Data):
+        if self.uniquify_channel_ids:
+            data.channels.id = np_string_prefix(f"{data.session.id}/", data.channels.id)
```
estefanysuarez, line 172: I believe, but I'm not 100% sure, that a more _torch_brain_aligned_ way of doing this would be using the class attribute `namespace_attributes` and overriding the method `apply_namespace()`, if I understand correctly what those are doing 😛

brainsets/utils/mne_utils.py
```diff
@@ -653,3 +653,37 @@ def _transpose_type_channels_mapping(
...
+    _check_mne_available("extract_annotations")
+    annotations = recording_data.annotations
+
+    if len(annotations) == 0:
+        warnings.warn("No annotations found in recording data")
+        return Interval(start=np.array([]), end=np.array([]))
+
+    start = np.array(annotations.onset)
+    extras = recording_data.annotations.extras
+    kwargs = {
```
estefanysuarez, line 689: Sometimes because of precision errors, it could happen that the start time of a trial/annotation is less than end time of the previous trial/annotation. Maybe verify that if by any chance this happens, then set the the end time of the previous trial to be the same as the start tim of the next trial. Something like this:

```
if start_time < end_times[-1]:
     # the previous trial goes into the next one
     # this happens because of numerical precision issues
      assert end_times[-1] - start_time < 0.1, (
          f"found overlap between trials: start time of trial i: {start_time}, end time of trial i-1: {end_times[-1]}"
      )

     # we can clip the end time of the last trial
     end_times[-1] = start_time
```

brainsets_pipelines/cmi_mipdb_2016/constants.py
estefanysuarez: Agree with Milo on this one!

brainsets_pipelines/cmi_mipdb_2016/eyetracking.py
```diff
@@ -0,0 +1,695 @@
+"""Parsing and synchronization of SMI iView eyetracking text exports.
+
+The public entry point is :func:`process_session`, which returns a
+``temporaldata.Data`` object containing up to four
+``IrregularTimeSeries``: *samples*, *fixations*, *saccades*, *blinks*.
+All timestamps are converted to the EEG timescale via a linear transform
+fitted from matched event codes.
+
+Synchronization strategy:
+1) Build an index of all ET files and their paradigm segments.
```
estefanysuarez, line 10: Super minor but before starting to use the ET acronym maybe define it once. So here this could be:
`1) Build an index of all eye tracking (ET) files and their paradigm segments.`

estefanysuarez: Or even in the very first line of this docstring, after the first time you mention the term "eye tracking"

brainsets_pipelines/cmi_mipdb_2016/eyetracking.py
estefanysuarez: Since the only public entry point seems to be `process_session()` and all other functions seem to be part (of this very specific) ET-EEG matching procedure, I would actually make all methods private and only keep `process_session()` public.

brainsets_pipelines/cmi_mipdb_2016/eyetracking.py
```diff
@@ -0,0 +1,695 @@
...
+2) Match the EEG session to the best ET file. Paradigms are grouped by
+   start code, then full event code sequences are compared via LCS to
+   resolve ambiguities (e.g. false starts that share the same start code).
+3) Align per-paradigm event sequences (LCS) to collect anchor pairs.
```
estefanysuarez, line 14: What does LCS refer to?

brainsets_pipelines/cmi_mipdb_2016/eyetracking.py
```diff
@@ -0,0 +1,695 @@
+"""Parsing and synchronization of SMI iView eyetracking text exports.
+
+The public entry point is :func:`process_session`, which returns a
+``temporaldata.Data`` object containing up to four
+``IrregularTimeSeries``: *samples*, *fixations*, *saccades*, *blinks*.
+All timestamps are converted to the EEG timescale via a linear transform
+fitted from matched event codes.
+
+Synchronization strategy:
```
estefanysuarez, line 9: If not too complicated, I think it might be worth to provide a brief description of the intricacies and nuances of the ET ---> EEG matching procedure for this specific dataset. This could help understanding the why for the complexity of the code.

brainsets_pipelines/cmi_mipdb_2016/pipeline.py
```diff
@@ -0,0 +1,558 @@
...
+            step=epoch_duration,
+            drop_short=True,
+        )
+        logging.info(
+            f"Behavior-relevant: {len(paradigm_epochs)} paradigm epochs "
+            f"of {epoch_duration}s"
+        )
+
+        if len(paradigm_epochs) < SPLIT_N_FOLDS:
+            logging.warning(
```
estefanysuarez, line 512: Throw the warning but instead of skipping return single-fold splits?

estefanysuarez (commented):
Hey @oumgharbi this is awesome! The ET stuff 😵‍💫 crazy. I think you did an amazing work there!

I left a few comments/suggestions here and there, but overall this already looks great!

brainsets/datasets/CmiMIPDB2016.py
vinamarora8: I think the "more correct" way would be to use the `MultiChannelDatasetMixin` which does the prefixing and also provides a `get_channel_ids()` method :D

brainsets_pipelines/cmi_mipdb_2016/constants.py
oumgharbi: I checked and they were always null values. But i will put them back for safe measure.

PR #83, opened 2026-02-19
fixing some minor issues with SleepEDF Dataset

milosobral:
* Some minor fixes to the Kemp Sleep EDF pipeline
 * Adding more splits to it for intersession and intersubject.

## Summary by CodeRabbit

* **New Features**
  * Deterministic 3‑fold assignment utility for subject- and session-aware splits.
  * Datasets now accept fold_number and fold_type (intrasession, intersubject, intersession) with validation and fold-aware sampling behavior.

* **Breaking Changes**
  * Dataset constructor parameters changed: use fold_number and fold_type instead of the prior split selector.
  * Channel metadata fields renamed from plurals to singulars (ids/types → id/type); channel extraction now returns a structured container.

* **Tests**
  * Updated and added tests covering fold assignment, dataset folding behavior, and channel metadata changes.

### DIFF
--- a/brainsets/utils/split.py
+++ b/brainsets/utils/split.py
@@ -335,12 +335,21 @@ def generate_string_kfold_assignment(
     val_ratio: float = 0.2,
     seed: int = 42,
 ) -> List[str]:
-    """Generate deterministic k-fold train/valid/test assignments for a string ID.
+    """Generate deterministic per-fold train/valid/test assignments for one ID.
 
-    This function performs hash-based assignment and deterministically assigns a
-    string identifier to train, valid, or test for each fold. The same `string_id`
-    with the same seed will
-    always receive the same assignments, allowing independent processing in parallel.
+    The assignment is independent for each fold index ``k``, but follows a
+    deterministic two-step rule:
+
+    1. Compute a global bucket from ``md5(f"{string_id}_{seed}") % n_folds``.
+       The fold whose index equals this bucket is labeled ``"test"``.
+    2. For every other fold, compute a fold-specific hash
+       ``md5(f"{string_id}_{seed}_{k}")`` and map it to ``[0, 1)``.
+       If that value is below ``val_ratio``, the fold is ``"valid"``,
+       otherwise it is ``"train"``.
+
+    As a result, each ``string_id`` appears in the test split for exactly one
+    fold and is never in test for the remaining folds. This makes the output
+    reproducible across runs and safe for parallel processing.
 
     Args
     ----
@@ -358,6 +367,7 @@ def generate_string_kfold_assignment(
     List[str]
         List of fold assignments where index ``k`` corresponds to fold ``k`` and
         each value is one of ``"train"``, ``"valid"``, or ``"test"``.
+        Exactly one entry is ``"test"``.
 
     Examples
     --------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Added normalization schemes for sex, age, and species in `SubjectDescription` ([#78](https://github.com/neuro-galaxy/brainsets/pull/78)).
 - Added generic data extraction helpers in `mne_utils` to handle MNE Raw objects ([#78](https://github.com/neuro-galaxy/brainsets/pull/78)).
 - Enriched `s3_utils` with additional functionalities to get data from public buckets ([#78](https://github.com/neuro-galaxy/brainsets/pull/78)).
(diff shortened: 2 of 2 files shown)


brainsets/datasets/KempSleepEDF2013.py
```diff
@@ -13,7 +13,10 @@ def __init__(
         recording_ids: Optional[list[str]] = None,
         transform: Optional[Callable] = None,
         uniquify_channel_ids: bool = True,
-        split_type: Optional[Literal["fold_0", "fold_1", "fold_2"]] = "fold_0",
+        fold_number: Optional[int] = 0,
+        fold_type: Literal[
```
estefanysuarez: ~~Just to draw your attention to this but maybe you are already aware. I can see that in the `AjileDataModule` on Foundry you used `split_type` instead of `fold_type`. Not sure which one is the most up-to-date~~

Just realized you are using now `fold_type` which makes more sense 😉

estefanysuarez (approved):
LGTM

brainsets/utils/split.py
```diff
@@ -439,3 +439,83 @@ def generate_train_valid_splits_one_epoch(
...
+            fold_str = f"{fold_base}_{k}"
+            fold_bytes = fold_str.encode("utf-8")
+            fold_hash_obj = hashlib.md5(fold_bytes)
+            fold_hash_int = int(fold_hash_obj.hexdigest(), 16)
+            normalized_hash = (fold_hash_int % 10000) / 10000.0
+            if normalized_hash < val_ratio:
+                assignments[f"{fold_prefix}{k}_assignment"] = "valid"
+            else:
+                assignments[f"{fold_prefix}{k}_assignment"] = "train"
+    return assignments
```
vinamarora8: This function is great! I think the API could be made more general by doing a kfold split given any string (doesn't have to be subject or session_ids).

```python
def generate_kfold_assignment_from_string(
    s: str,
    n_folds: int = 3,
    val_ratio: float = 0.2,
    seed: int = 42,
) -> dict[str, str]:
```

You could send in `f"session_fold_{subject.id}_{session.id}"` etc. into this function. What do you think?

brainsets/datasets/KempSleepEDF2013.py
```diff
@@ -13,7 +13,10 @@ def __init__(
...
+            "intrasession", "intersubject", "intersession"
+        ] = "intrasession",
         dirname: str = "kemp_sleep_edf_2013",
         **kwargs,
     ):
```
vinamarora8: Docstring for the fold types would be great! Actually docstring for the entire dataset would be great too. But can be another PR if you do not have the time.

brainsets/utils/mne_utils.py
```diff
@@ -168,8 +167,8 @@ def extract_psg_signal(raw_psg: "mne.io.Raw") -> Tuple[RegularTimeSeries, ArrayD
     )
 
     channels = ArrayDict(
-        ch_id=np.array([ch["ch_id"] for ch in channel_meta], dtype="U"),
-        ch_type=np.array([ch["ch_type"] for ch in channel_meta], dtype="U"),
+        ids=np.array([ch["ids"] for ch in channel_meta], dtype="U"),
+        types=np.array([ch["types"] for ch in channel_meta], dtype="U"),
     )
```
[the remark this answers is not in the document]
milosobral: Oh ok I thought it was the other way around but that's fine by me!

brainsets/datasets/KempSleepEDF2013.py
milosobral: No that's a good point

brainsets/utils/split.py
```diff
@@ -472,32 +468,29 @@ def generate_subject_kfold_assignment(
...
-    >>> assignments_sess
-    {'session_fold_0_assignment': 'valid', 'session_fold_1_assignment': 'train',
-     'session_fold_2_assignment': 'test'}
+    {'fold_0_assignment': 'train', 'fold_1_assignment': 'test',
+     'fold_2_assignment': 'train'}
+
+    >>> generate_string_kfold_assignment("sub-01_ses-01", n_folds=3)
+    {'fold_0_assignment': 'valid', 'fold_1_assignment': 'train',
+     'fold_2_assignment': 'test'}
     """
```
vinamarora8: Do you think maybe the output should be just a list? e.g.
```
>>> generate_string_kfold_assignment("sub-01_ses-01", n_folds=3)
['valid', 'train', 'test']
```
The key-value system seems redundant..

vinamarora8 (commented):
Leaving a couple nit comments (sorry :3)

vinamarora8 (commented):
Left a minor comment. Can you please also add these changes to CHANGELOG.md? I guess both things: changes regarding sleepEDF and the new util methods. Ready to merge after these!

Issue #94, opened 2026-03-03
Feature: Make stratification optional for generating folds


oumgharbi:
The current `generate_stratified_folds` requires categorical label for stratification, preventing use for non-stratified datasets.
https://github.com/neuro-galaxy/brainsets/blob/6317860c66fdafd56ceb0fb04e61c35589af1838/brainsets/utils/split.py#L242-L248

**Expected Behavior**  
Users should generate random K-folds without stratification when no label is provided (`stratify_by=None`).

PR #95, opened 2026-03-03
Fold generation to support optional stratification

oumgharbi:
Adressing issue #94.
Small PR to modify the util to generate folds, making stratification optional and used when a label is provided. 
Otherwise, random K-folds are generated without stratification. 
`generate_stratified_folds` is updated as `generate_folds`, with the `stratify_by` argument used to differentiate logics.

- Tests are updated for both use cases. 
- Kemp_sleep_edf_2013 is updated to use the new util. 

## Release Notes

* **New Features**
  * Fold generation now supports optional stratification, enabling both stratified and non-stratified splitting modes based on user preference.
  * Added validation for fold count, validation ratio, and attribute presence to catch configuration errors.

* **API Changes**
  * Fold generation function renamed with updated parameter signature; stratification is now optional rather than required.

### DIFF
--- a/brainsets/utils/split.py
+++ b/brainsets/utils/split.py
@@ -239,92 +239,133 @@ def _create_interval_split(intervals: Interval, indices: np.ndarray) -> Interval
     return split
 
 
-def generate_stratified_folds(
+def generate_folds(
     intervals: Interval,
-    stratify_by: str,
     n_folds: int = 5,
     val_ratio: float = 0.2,
     seed: int = 42,
+    stratify_by: Optional[str] = None,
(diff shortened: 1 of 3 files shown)


vinamarora8 (commented):
Hi @oumgharbi, I like the idea of having a `generate_folds` without stratification. However, I see that the logic between the two is not quite reusable, and in your code you had to basically have an `if/else` to choose between stratified logic or non-stratified logic.

In this case, I think we might as well create a new function `generate_folds`, and keep the old `generate_stratified_folds` as it is.

Issue #97, opened 2026-03-06
Expose `--download-only` flag in the CLI


oumgharbi:
Brainsets CLI currently allows to prepare a brainset, by downloading raw data and processing it. These 2 steps are usually done together. 

Separating download and processing would be useful when working with limited computational resources -- for example, downloading data on a login node and deferring processing to a compute node.

The `--download-only` flag is actually already implemented in the runner and pipeline layers. However, it is not discoverable: it doesn't appear in `brainsets prepare --help`.

 #### Proposed change:
Add `--download-only` as an explicit `@click.option` in `cli_prepare.py`.

PR #98, opened 2026-03-06
Expose --download-only option in CLI

oumgharbi:
Adress issue #97.

`--download-only` option is fully implemented but not discoverable at CLI level. 

Minimal PR to add `--download-only` as an explicit `@click.option` in `cli_prepare.py` so that:
- It shows up in `brainsets prepare --help`
- It is properly documented with a help string
- It is explicitly forwarded to the runner subprocess command

## Summary by CodeRabbit

* **New Features**
  * Added a --download-only flag to the prepare CLI, allowing users to download required resources without running the full pipeline.

* **Documentation**
  * Updated CLI help text and usage examples to document the new flag.

* **Tests**
  * Added a test to verify the --download-only flag is forwarded to the execution step.

* **Chore**
  * Added changelog entry describing the new flag.


vinamarora8 (approved):
lgtm!

PR #99, opened 2026-03-09
Add all HBN releases

oumgharbi:
New pipeline `cmi_hbn_2025 ` that prepares the [Healthy Brain Network (HBN)](https://openneuro.org/datasets/ds005505/versions/1.0.1) EEG dataset across all 11 OpenNeuro releases in one place. It reuses the `OpenNeuroEEGPipeline` intorduced in #75 and groups all 11 datasets into one via minimal overrides.

### Features:
- One `pipeline.py` for all 11 releases.
- The pipeline provides the same channels and processing for all releases.
- Release selection via argument : `--release`
     - `brainsets prepare cmi_hbn_2025` → prepare all 11 releases
     - `brainsets prepare cmi_hbn_2025 --release 3` → prepare only release 3 (ds005507)
- Per-release layout: creates subdirs for each release for raw and processed data
     - Raw: `.../cmi_hbn_2025/R1/`, `R2/`, …, `R11/` (BIDS + `dataset_description.json` per release).
     - Processed: `.../cmi_hbn_2025/R1/*.h5`, `R2/*.h5`, etc.
- Error handling (HBN pipeline only)
     - Catch invalid `--release`
     - When using both `--release` and `--single`, run a cross check that the selected session is in the selected release, else skip.

### DIFF
--- a/brainsets_pipelines/cmi_hbn_2025/pipeline.py
+++ b/brainsets_pipelines/cmi_hbn_2025/pipeline.py
@@ -1,0 +1,82 @@
+# /// brainset-pipeline
+# python-version = "3.11"
+# dependencies = [
+#   "mne==1.11.0",
+#   "mne-bids==0.18",
+#   "boto3>=1.42.32",
+#   "requests==2.32.5",
+# ]
+# ///
+
+from argparse import ArgumentParser
+
+import pandas as pd
+
+from brainsets.utils.openneuro import OpenNeuroEEGPipeline
+from brainsets.utils.openneuro.pipeline import _openneuro_parser
+
+RELEASES = {
+    1: "ds005505",
+    2: "ds005506",
+    3: "ds005507",
+    4: "ds005508",
+    5: "ds005509",
+    6: "ds005510",
+    7: "ds005511",
+    8: "ds005512",
+    9: "ds005514",
+    10: "ds005515",
+    11: "ds005516",
+}
+
+MODALITY_CHANNELS = {"EEG": [f"E{i}" for i in range(1, 129)] + ["Cz"]}
+
+parser = ArgumentParser(parents=[_openneuro_parser], add_help=False)
+parser.add_argument("--release", type=int, choices=range(1, 12), default=None)
+
+
+class Pipeline(OpenNeuroEEGPipeline):
+    brainset_id = "cmi_hbn_2025"
+    dataset_id = "ds005505"
+    description = (
+        "Healthy Brain Network (HBN) EEG dataset combining all 11 data "
+        "releases. Contains recordings from participants performing various "
+        "passive and active tasks including resting state, movie watching, "
+        "and cognitive tasks."
+    )
+    MODALITY_CHANNELS = MODALITY_CHANNELS
+    parser = parser
+
+    @classmethod
+    def get_manifest(cls, raw_dir, args):
+        if args and args.release is not None:
+            releases = {args.release: RELEASES[args.release]}
+        else:
+            releases = RELEASES
+
+        all_manifests = []
+        for release_id, dataset_id in releases.items():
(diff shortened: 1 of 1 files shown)


brainsets_pipelines/cmi_hbn_2025/pipeline.py
```diff
@@ -0,0 +1,94 @@
...
+#   "boto3>=1.42.32",
+#   "requests==2.32.5",
+# ]
+# ///
+
+from argparse import ArgumentParser
+
+import pandas as pd
+
+from brainsets.utils.openneuro import OpenNeuroEEGPipeline
```
estefanysuarez, line 15: After a short discussion with Vinam, we agreed that for now we would only have a single `OpenNeuroPipeline` class, so `OpenNeuroIEEGPipeline` and `OpenNeuroEEGPipeline` no longer exist 😔 

The only part impacted in this PR would be in the definition of the Pipeline (i.e., adding `modality` as class attribute):

```
class Pipeline(OpenNeuroEEGPipeline):
    modality = "eeg"
    brainset_id = "cmi_hbn_2025"
    dataset_id = "ds005505"
    description = (
        "Healthy Brain Network (HBN) EEG dataset combining all 11 data "
        "releases. Contains recordings from participants performing various "
        "passive and active tasks including resting state, movie watching, "
        "and cognitive tasks."
    )
    MODALITY_CHANNELS = MODALITY_CHANNELS
    parser = parser
```

brainsets_pipelines/cmi_hbn_2025/pipeline.py
```diff
@@ -0,0 +1,94 @@
...
+class Pipeline(OpenNeuroEEGPipeline):
+    brainset_id = "cmi_hbn_2025"
+    dataset_id = "ds005505"
+    description = (
+        "Healthy Brain Network (HBN) EEG dataset combining all 11 data "
+        "releases. Contains recordings from participants performing various "
+        "passive and active tasks including resting state, movie watching, "
+        "and cognitive tasks."
+    )
+    MODALITY_CHANNELS = MODALITY_CHANNELS
```
estefanysuarez, line 52: Either rename this attribute as TYPE_CHANNELS_REMAPPING (which would be my recommendation and it's easier) or modify the `get_type_channels_remapping()` method to return `MODALITY_CHANNELS` instead of the default `self.TYPE_CHANNELS_REMAPPING`, meaning:

   ```
 def get_type_channels_remapping(
        self,
        recording_id: str | None = None,
    ) -> Optional[dict[str, list[str]]]:
        """Return channel type remapping for a given recording.

        Override this method to provide per-recording channel type remappings.
        The default implementation returns the class-level TYPE_CHANNELS_REMAPPING attribute.

        Args:
            recording_id: The recording identifier

        Returns:
            Mapping from channel type to channel name list, or ``None``.
        """
        return self.MODALITY_CHANNELS

```

File: brainsets_pipelines/cmi_hbn_2025/pipeline.py [hunk 3]
```diff
@@ -0,0 +1,94 @@
...
+    @classmethod
+    def get_manifest(cls, raw_dir, args):
+        if args and args.release is not None:
+            if args.release not in RELEASES:
+                raise ValueError(
+                    f"Invalid choice: release must be one of 1-11, got {args.release}."
+                )
+            releases = {args.release: RELEASES[args.release]}
+        else:
+            releases = RELEASES
```
estefanysuarez, line 64: Do you think it might be worth to allow users to provide either the number of the realease OR the ID of the dataset? For example, instead of `5` do `ds005509`?

estefanysuarez (commented):
I left a few comments. Most of them you would have realized them sooner or later after merging the OpenNeuro PR anyways 😁

PR #117, opened 2026-04-02
chore!: Cleanup dandi_utils.py and add docs

vinamarora8:
- Added `_check_dandi_available()` like in other util modules
- Added docstrings
- Rm `extract_metadata_from_nwb` as it is not used anywhere and was undocumented (plus, i don't see it being used in the future)

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed an unused NWB metadata extraction capability; migrate to remaining extraction workflows.

* **Improvements**
  * Standardized error messaging when an optional external library is missing.
  * Strengthened public interfaces with explicit type annotations and clearer parameter/return docs.
  * Improved subject/species normalization and clarified supported recording types for spike extraction.
  * Streamlined download and NWB asset-listing behavior.


brainsets/utils/dandi_utils.py
```diff
@@ -11,22 +12,40 @@
...
+
+    Returns:
+        A :obj:`SubjectDescription <brainsets.descriptions.SubjectDescription>`
     """
+
+    # DANDI has requirements for metadata included in `subject`
+    # - subject_id: A subject identifier must be provided.
+    # - species: either a latin binomial or NCBI taxonomic identifier.
+    # - sex: must be "M", "F", "O" (other), or "U" (unknown).
+    # - date_of_birth or age: this does not appear to be enforced, so will be skipped.
```
milosobral, line 58: Is there a reason for not including that in the docstring? I feel like it could be useful no?

milosobral (dismissed):
Small question but otherwise looks good!

oumgharbi (dismissed):
LGTM!

brainsets/utils/dandi_utils.py
vinamarora8: I just thought it is more info than needed in a docstring. But happy to add it in there if you think otherwise!

oumgharbi:
@vinamarora8 Can you please remove the `extract_metadata_from_nwb` reference from documentation also ?

vinamarora8:
@oumgharbi done!

oumgharbi (approved):
Thank you @vinamarora8 🙂

PR #119, opened 2026-04-02
chore!: Remove mat_utils.py (legacy code)

vinamarora8:
This is an old module that was used to process very early versions of PerichMiller dataset. 
Since it is not needed anymore, we should remove it from the public package.

## Summary by CodeRabbit

* **Removed**
  * Removed a legacy MATLAB file utility module and its public API from the package, and deleted the corresponding documentation entry so the utility is no longer exposed or documented.


nandahkrishna (commented):
Hey @vinamarora8, @TomGeorge1234 and I were planning to add a hippocampal dataset soon where this `loadmat` function could be really useful. We're planning to have that out as a PR soon. Any chance we could retain this?

(WIP branch: https://github.com/neuro-galaxy/brainsets/tree/vollan_moser_alternating_2025)

milosobral (dismissed):

vinamarora8:
@nandahkrishna were you referring to the dataset in #127? If so, turns out `mat_utils` was not needed there and we can safely remove it.

nandahkrishna:
Yes we went ahead without using `mat_utils`, it was fine with some extra args passed to `sio.loadmat`. This should be good to merge!

nandahkrishna (approved):

PR #126, opened 2026-04-09
refactor!: Move split_*_epochs to allen vc ophys pipeline

vinamarora8:
The `split_*_epochs` methods were specifically designed for the visual presentation trial structure of Allen VC datasets. These are otherwise not well documented and not tested. So, IMO, they should not be provided as part of the public API at the moment.

This PR:
- Removes `split_*_epochs` and `generate_train_val_test_splits` from `brainsets.units.split`
- These are added to the `allen_visual_coding_ophys_2016` pipeline directory, which is the only place they are used.


AlexandreAndr (approved):
Looks good to me!

PR #125, opened 2026-04-09
feat: BrainsetPipeline runner creates raw and processed dirs

vinamarora8:
Closes #63

## Summary by CodeRabbit

* **New Features**
  * Pipeline runner now automatically creates raw and processed output directories during initialization, simplifying pipeline setup and removing the need for manual directory creation.

* **Documentation**
  * Updated pipeline creation documentation with formatting improvements.

PR #127, opened 2026-04-12
Add `vollan_moser_alternating_2025` dataset

TomGeorge1234:
## Add Vollan/Moser grid cell brainset

This is a pretty gnarly dataset brimming with some of the clearest examples of grid cells, head direction cells and place cells I've ever seen. Plus I think the first hippocampus/MEC brainset yet which should broaden the appeal quite substantially. It's a very recently released dataset , published in Nature, originally used to study theta sequences. The raw data is publicly hosted on [EBRAINs](https://search.kg.ebrains.eu/instances/4080b78d-edc5-4ae4-8144-7f6de79930ea)

In summary:
* 51 recordings (42 awake navigation, 9 asleep) across 19 rats. 
* 45,821 units for HPC or MEC or both (average of 899 per recording, which is ridiculous) 
* 63.3 hours (34.3 after they filter out speeds < 5 cm/s) 
* 215 million spikes. 
* Raw size ~16.4GB, processed size ~4.5GB

There is some nice metadata too, for example the identity of the cells (HPC vs MEC), the kilosort label ("good", "bad", "mua") and whether they have been identified as grid cells or not.

Included behaviours are x y z position as well as head direction. They also include LFP theta phase which makes this a great dataset for studying temporal coding. 

## Summary

- Add `vollan_moser_alternating_2025` brainset: Neuropixels recordings from MEC and hippocampus in freely moving rats performing spatial navigation tasks (open field, linear track, M-maze, wagon wheel) and during sleep, from [Vollan et al. (2025)](https://www.nature.com/articles/s41586-024-08527-1)
- Pipeline downloads 51 `.mat` files from the EBRAINS data proxy and processes them into standardized HDF5 format (16.4 GB raw -> 4.5 GB processed)
- Includes 17 subjects, 42 navigation sessions (with behavior: position, head direction, theta, decoded internal direction) and 9 sleep sessions (SWS + REM epochs)
- Adds new taxonomy entries: `Species.RATTUS_NORVEGICUS`, and 5 new navigation `Task` types (open field, linear track, M-maze, wagon wheel, sleep)
- Adds `VollanMoserAlternating2025` dataset class with `SpikingDatasetMixin`
- Adds brainset card to documentation and entry to README table

### Visualisation 

Here's a little picture of the beautiful grid cells from the open-field recording `of_25691_2` (note this recording has 641 units in total but only 173 "grid cells"...not all MEC cells are grid cells. I'm only showing the grid cells here)

gridcells

## Summary by CodeRabbit

* **New Features**
  * Added the vollan_moser_alternating_2025 brainset with download/process CLI and processing pipeline; exposed at package level.
  * Added Norway rat species and several new Task types (navigation variants and sleep).

* **Documentation**
  * README updated with new brainset entry (raw/processed sizes; Brainset Card raw-data TBA).
  * Glossary/docs added describing the brainset and download/preparation instructions.

* **Tests**
  * Added unit tests for domain construction, behavior extraction, and unit/spike extraction.


brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
```diff
@@ -0,0 +1,580 @@
...
+    if sws_times.size > 0:
+        sws_times = np.atleast_2d(sws_times)
+        all_starts.append(sws_times[:, 0])
+        all_ends.append(sws_times[:, 1])
+
+    if rem_times.size > 0:
+        rem_times = np.atleast_2d(rem_times)
+        all_starts.append(rem_times[:, 0])
+        all_ends.append(rem_times[:, 1])
+
```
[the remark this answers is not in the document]
TomGeorge1234: true but the dataset is fixed and this doesn't happen in practice

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
```diff
@@ -0,0 +1,580 @@
...
+        session_description = SessionDescription(
+            id=session_id,
+            recording_date=datetime.datetime(2025, 1, 1),
+            task=task,
+        )
+
+        device_description = DeviceDescription(
+            id=f"{animal_id}_neuropixels",
+            recording_tech=RecordingTech.NEUROPIXELS_SPIKES,
+        )
```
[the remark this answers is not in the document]
TomGeorge1234: done

docs/source/glossary/brainsets.rst
```diff
@@ -887,3 +887,146 @@ perich_miller_population_2018
...
+                   <button class="cite-button" data-doi="10.25493/R5FR-EDG">Cite</button>
+                   <div class="citation-popup" id="popup-10.25493/R5FR-EDG" style="display:none;">
+                     <div class="citation-section">
+                       <h4>BibTeX</h4>
+                       <div class="citation-content">
+                         <pre>@article{vollan2025theta,
+    title={Left-right-alternating theta sweeps in the entorhinal-hippocampal spatial map},
+    author={Vollan, Abraham Z and Gardner, Richard J and Moser, May-Britt and Moser, Edvard I},
+    year={2025}
+    }
```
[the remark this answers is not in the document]
TomGeorge1234: fixed

docs/source/glossary/brainsets.rst
```diff
@@ -887,3 +887,146 @@ perich_miller_population_2018
...
+           <td>19</td>
+       </tr>
+       <tr>
+           <td><strong>Species</strong></td>
+           <td>Rattus norvegicus</td>
+       </tr>
+        <tr>
+           <td><strong>Number of recordings</strong></td>
+           <td>51 (31 Open Field, 7 Linear Track, 2 Wagon Wheel, 1 M-Maze, 1 Novel Open Field, 9 Sleep)</td>
+       </tr>
```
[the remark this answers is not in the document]
TomGeorge1234: fixed (19 was correct)

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
```diff
@@ -0,0 +1,580 @@
...
+        brainset_description = BrainsetDescription(
+            id="vollan_moser_alternating_2025",
+            origin_version="sharing_v4",
+            derived_version="1.0.0",
+            source="https://doi.org/10.6084/m9.figshare.28443963",
+            description="Neuropixels recordings from MEC and hippocampus in rats "
+            "performing spatial navigation tasks (open field, linear track, M-maze, "
+            "wagon wheel) and during sleep. Includes grid cells, head direction cells, "
+            "and other spatially tuned neurons.",
+        )
```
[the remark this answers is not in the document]
TomGeorge1234: set to EBRAINs all over

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
```diff
@@ -0,0 +1,580 @@
...
+            sex=Sex.UNKNOWN,
+        )
+
+        task = SESSION_TYPE_TO_TASK[session_type]
+
+        session_description = SessionDescription(
+            id=session_id,
+            recording_date=datetime.datetime(2025, 1, 1),
+            task=task,
+        )
```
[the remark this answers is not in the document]
TomGeorge1234: true recording dates are unavailable and so this placeholder is consistent with the paper publication year

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
```diff
@@ -0,0 +1,580 @@
...
+
+        # Handle empty regions (numpy array with size 0 or empty list)
+        if isinstance(region_units, np.ndarray) and region_units.size == 0:
+            continue
+        if isinstance(region_units, list) and len(region_units) == 0:
+            continue
+
+        # With simplify_cells=True, this is a list of dicts
+        if not isinstance(region_units, list):
+            continue
```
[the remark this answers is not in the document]
TomGeorge1234: fixed in pipeline.py

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
```diff
@@ -0,0 +1,580 @@
...
+        all_ends.append(sws_times[:, 1])
+
+    if rem_times.size > 0:
+        rem_times = np.atleast_2d(rem_times)
+        all_starts.append(rem_times[:, 0])
+        all_ends.append(rem_times[:, 1])
+
+    starts = np.concatenate(all_starts).astype(np.float64)
+    ends = np.concatenate(all_ends).astype(np.float64)
+
```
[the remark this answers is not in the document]
TomGeorge1234: added tests covering these helper functions. all tests pass.

tests/test_vollan_pipeline.py
```diff
@@ -0,0 +1,231 @@
...
+
+def test_build_domain_single_sample():
+    """A single timestamp should produce one interval of length dt_nominal."""
+    t = np.array([5.0])
+    # With a single sample there's no diff, but the function uses np.median(np.diff(t))
+    # which will be empty — let's check it doesn't crash.
+    # Actually np.median of empty array returns nan, so let's test with 2 samples.
+    t = np.array([5.0, 5.01])
+    domain = build_domain_from_timestamps(t)
+    assert len(domain.start) == 1
```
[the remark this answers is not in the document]
TomGeorge1234: resolved in latest commit

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
```diff
@@ -0,0 +1,583 @@
...
+    # Segment boundaries: each gap splits the timeseries
+    gap_indices = np.where(gap_mask)[0]
+
+    starts = np.empty(len(gap_indices) + 1, dtype=np.float64)
+    ends = np.empty(len(gap_indices) + 1, dtype=np.float64)
+
+    starts[0] = t[0]
+    ends[-1] = t[-1] + dt_nominal  # extend last segment by one sample
+
```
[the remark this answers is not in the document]
TomGeorge1234: resolved in latest commit

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
```diff
@@ -0,0 +1,591 @@
...
+        for i, u in enumerate(region_units):
+            unit_meta.append(
+                {
+                    "id": f"{location}_{i}",
+                    "location": location,
+                    "probe_id": int(u["probeId"]),
+                    "shank": int(u["shank"]),
+                    "shank_pos": float(u["shankPos"]),
+                    "mean_rate": float(u["meanRate"]),
+                    "is_grid": int(u.get("isGrid", 0)),
```
vinamarora8, line 582: Can `u["ks2Label"]` also be added to this please?

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
```diff
@@ -0,0 +1,591 @@
...
+                    "fname": fname,
+                }
+            )
+
+        manifest = pd.DataFrame(manifest_list).set_index("session_id")
+        return manifest
+
+    def download(self, manifest_item):
+        self.raw_dir.mkdir(parents=True, exist_ok=True)
+
```
vinamarora8: Not needed anymore after #125

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
```diff
@@ -0,0 +1,591 @@
...
+            subject=subject,
+            session=session_description,
+            device=device_description,
+            # neural activity
+            spikes=spikes,
+            units=units,
+            # behavior
+            behavior=behavior,
+            # metadata
+            probe_channels=probe_channels,
```
vinamarora8, line 306: Curious: I presume probe_channels includes metadata about the probe and its channels. How would you know which unit came from which channel? I see there is probe_id and shank in `units` but not channel_id.

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
vinamarora8, line 582: Maybe: proposing `shank` to be renamed as `shank_id` to be consistent with `probe_id`.

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
```diff
@@ -0,0 +1,591 @@
...
+def extract_navigation_units_and_spikes(ds, domain):
+    """Extract units metadata and spikes from Dsession.units (MEC + HC)."""
+    units_struct = ds["units"]
+    unit_meta = []
+    spike_timestamps_list = []
+    spike_unit_index_list = []
+    unit_idx = 0
+
+    for location in ["mec", "hc"]:
+        region_units = units_struct[location]
```
vinamarora8, line 557: Are you certain there are no more keys in `units_struct`? If so, can you write a small comment above this line.

brainsets/datasets/__init__.py
```diff
@@ -33,3 +33,4 @@
 from .AllenVisualCodingOphys2016 import AllenVisualCodingOphys2016
 from .KempSleepEDF2013 import KempSleepEDF2013
 from .Neuroprobe2025 import Neuroprobe2025
+from .VollanMoserAlternating2025 import VollanMoserAlternating2025
```
vinamarora8, line 37: Can you also add the dataset to `_electrophysiology_dataset` list above? This will make its docstring visible in the docs

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
```diff
@@ -0,0 +1,591 @@
...
+            self._process_navigation(fpath, session_id, session_type)
+        elif fpath.parent.name == "sleep":
+            session_id = f"sleep_{fpath.stem}"
+            self._process_sleep(fpath, session_id)
+        else:
+            raise ValueError(f"Cannot infer session type from path: {fpath}")
+
+    def _process_navigation(self, fpath, session_id, session_type):
+        self.processed_dir.mkdir(exist_ok=True, parents=True)
+
```
vinamarora8: Not needed anymore after #125

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
```diff
@@ -0,0 +1,591 @@
...
+        brainset_description = BrainsetDescription(
+            id="vollan_moser_alternating_2025",
+            origin_version="sharing_v4",
+            derived_version="1.0.0",
+            source="https://doi.org/10.25493/R5FR-EDG",
+            description="Neuropixels recordings from MEC and hippocampus in rats "
+            "performing spatial navigation tasks (open field, linear track, M-maze, "
+            "wagon wheel) and during sleep. Includes grid cells, head direction cells, "
+            "and other spatially tuned neurons.",
+        )
```
vinamarora8: Can you define this outside this method near the top of this file. Then it won't have to be repeated in `_process_sleep`, making it easier to keep consistent if there are any changes in the future.

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
```diff
@@ -0,0 +1,591 @@
...
+# Helper functions
+# ---------------------------------------------------------------------------
+
+
+def build_domain_from_timestamps(t):
+    """Build an Interval domain from speed-filtered timestamps.
+
+    Detects gaps in the timestamp array (where dt > 2x the nominal sampling
+    interval) and creates separate intervals for each contiguous segment.
+    """
```
vinamarora8, line 415: This seems a bit different from other datasets, so some more context in the docstring would be great for people trying to understand the pipeline! E.g., what is `ds["t"]` (the arg to this function)?

vinamarora8 (commented):
Mostly clean! Leaving some comments to improve things a bit, mostly for maintainability

vinamarora8:
Also, thanks @TomGeorge1234 for taking this up!

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
TomGeorge1234: Done!

TomGeorge1234: Done!

brainsets/datasets/__init__.py
TomGeorge1234: Done!

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
TomGeorge1234: Removed ✅

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
TomGeorge1234: Removed ✅

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
TomGeorge1234: `ds['t']` is speed-filtered timestamps (times when animal is < 5 cm/s are removed by the original data creators). I've added a comment explaining.

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
TomGeorge1234: Definitely only two. Comment added ✅

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
TomGeorge1234: Every other pipeline in the repo seems to define it inline in the process method.  Extracting it here would be inconsistent. Happy to move if you feel strongly otherwise I think this should be done repo-wide in another PR.

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
TomGeorge1234: My understanding is that the raw MAT files don't provide a direct unit→channel mapping. Each unit has probe_id, shank_id, and shank_pos (a position along the shank in microns). Each channel in probe_channels has probe_id, shank_id, and y_um (its y-coordinate on the shank).

So to associate a unit with its channel, you'd join on probe_id + shank_id and then find the channel whose y_um is closest to the unit's shank_pos. It's not an exact key lookup — it's a nearest-neighbor match on the probe geometry.

I've added a comment explaining this ✅

TomGeorge1234:
@vinamarora8 I think I resolved all your comments. Thanks for these! Just one for you to check then its g2g as far as I'm concerned.

All tests pass and reprocessing the brainiest still works fine with the latest changes.

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
vinamarora8: I see what you're pointing to. I'm thinking of this situation as a special case because this definition has to be done _twice_. Once in `_process_navigation` and once in `_process_sleep`. This has not been the case in other pipelines I believe.

My concern is: in the future let's say someone makes an edit here that warrants a version increment. While incrementing the `derived_version`, it's possible they do it in only one method and forget to do it the other. I'm just trying to avoid that situation and I think it's okay to break consistency for that.

In fact, the consistency has already be broken by Neuroprobe [here](https://github.com/neuro-galaxy/brainsets/blob/4b8b46a4e6dc7bb3651822027566bd01b3479a5a/brainsets_pipelines/neuroprobe_2025/pipeline.py#L310) for a separate good reason. So I don't think maintaining consistency about it is a big deal. Pipelines just need to function well and be maintainable :)

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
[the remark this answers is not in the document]
vinamarora8: Can you use the UNIX epoch in this case:  `datetime.datetime(1970, 1, 1, tzinfo=datetime.timezone.utc)`

Pattern established in the `utils/mne_utils.py` file. 
If it's 1970, it's easier to guess that the date is meaningless.

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
vinamarora8: perfect!

vinamarora8 (commented):
Thanks for the edits! Some more discussion below

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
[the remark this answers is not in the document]
TomGeorge1234: fair enough. done and left a comment too clarifying its a placeholder

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
TomGeorge1234: Ok yep, seems fair enough, I've moved it out and up to a new variable called `BRAINSET_DESCRIPTION` should be had for anyone to miss

TomGeorge1234:
@vinamarora8 You're right on those two. All fixed now

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
```diff
@@ -0,0 +1,589 @@
...
+        return manifest
+
+    def download(self, manifest_item):
+        # Each file is downloaded individually from the ebrains data-proxy API
+        fpath = self.raw_dir / manifest_item.fname
+        if fpath.exists() and not self.args.redownload:
+            return fpath
+
+        fpath.parent.mkdir(parents=True, exist_ok=True)
+        url = f"{BUCKET_URL}/sharing_v4/{manifest_item.fname}"
```
milosobral: It might be better to include this `sharing_v4` as a scriptwide constant in case there are changes in the future

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
```diff
@@ -0,0 +1,589 @@
...
+        return fpath
+
+    def process(self, fpath):
+        fpath = Path(fpath)
+        # Infer data category and session type from file path
+        # Navigation: .../navigation/<session_type>/<animal>_<rec>.mat
+        # Sleep: .../sleep/<animal>_<rec>.mat
+        if fpath.parent.parent.name == "navigation":
+            session_type = fpath.parent.name  # of, mmaze, lt, ww, of_novel
+            session_id = f"{session_type}_{fpath.stem}"
```
milosobral: Doens't really affect the functionality but I think you already compute those session IDs based on the session_type and stem in the get_manifest(). Instead of redoing that here, you could pass it through `download` or simply remove it from the manifest if it is never used

brainsets/datasets/VollanMoserAlternating2025.py
```diff
@@ -0,0 +1,71 @@
...
+    .. admonition:: Preprocessing
+
+        To download and prepare this dataset, run
+        ``brainsets prepare vollan_moser_alternating_2025``.
+
+    **Tasks:** Open Field, Linear Track, M-Maze, Wagon Wheel, Sleep
+
+    **Brain Regions:** MEC, Hippocampus
+
+    **Dataset Statistics**
```
milosobral, line 134: Given that Navigation and Sleep sessions end up with quite significant differences in the data they contain, a more detailed description of what data is where in each type of session would be beneficial here.

milosobral (commented):
Minor comments

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
TomGeorge1234: Cool 🫘

TomGeorge1234: ((ps. I doubt there will be changes neuroscientist aren't that diligent ;)))

brainsets/datasets/VollanMoserAlternating2025.py
TomGeorge1234: Yeah, they should be seen as totally different datasets really. Not much we can do to "unify" them but best to at least include it and clarify the difference. Comment added in `f43f57af`

brainsets_pipelines/vollan_moser_alternating_2025/pipeline.py
TomGeorge1234: Done, session id now passed through download and not recomputed. Good spot!

TomGeorge1234:
I think we might be good to merge this now...the only possibility I see is that in the future we may want to add additional data fields available in the raw which we don't currently process (its quite a rich dataset). We can always do that in a future PR. For now the basics are all in there. 

@nandahkrishna anything you want to chip in here?

nandahkrishna (dismissed):
I looked through it and didn't really spot anything amiss, I think we can add features in future PRs and fix forward if there are any issues we spot!

milosobral (dismissed):

brainsets/datasets/VollanMoserAlternating2025.py
```diff
@@ -0,0 +1,270 @@
...
+        "lt": lambda: list(RECORDING_IDS.navigation.lt),
+        "mmaze": lambda: list(RECORDING_IDS.navigation.mmaze),
+        "ww": lambda: list(RECORDING_IDS.navigation.ww),
+        "of_novel": lambda: list(RECORDING_IDS.navigation.of_novel),
+    }
+
+    def __init__(
+        self,
+        root: str,
+        recording_ids=None,
```
milosobral: Could you add a type hint here please? 🫰

milosobral (commented):
Very small comment but otherwise looks good!

vinamarora8 (commented):
Looks good! The detailed docstring is great. Final thing from me, forgot to ask previously: Can you please add this pipeline to `.github/workflows/pipeline_testing.yml`. We test single session preprocessing there to ensure future changes to core utils don't break any pipelines catastrophically.

vinamarora8:
Re my previous comment: You could also add 1 session each from both navigation and sleep since they follow different paths

brainsets/datasets/VollanMoserAlternating2025.py
TomGeorge1234: ✅

TomGeorge1234:
@vinamarora8 thanks for pointing that out, I hadn't seen those tests (its reassuring they happen). Added now ✅

milosobral (approved):
Looks good!

vinamarora8 (approved):
lgtm!

vinamarora8:
Thanks @TomGeorge1234! Great addition

PR #130, opened 2026-04-15
docs: update `brainsets config` in CLI reference

vinamarora8:
Add changes from #122  to the CLI reference doc


nandahkrishna (approved):
Thanks Vinam!

PR #137, opened 2026-04-22
feat(cli): make `--list` and `-s`/`--single` public

vinamarora8:
Simple change. `--list` and `-s`/`--single` have proven to be useful CLI features for the `brainsets prepare` command, especially for development and debugging phase. These were not "public", in the sense that they were not exposed by the `brainsets prepare` command, and hence would not show up in `brainsets prepare --help`. 

This PR 
- Exposes both options/flags from `brainsets prepare`. 
- It also adds them to the CLI reference doc page.


oumgharbi (approved):
Great, I find these options very useful for some tests without having to go through the entire dataset. LGTM!

PR #139, opened 2026-05-15
chore: Remove brainset cards

divyansha1115:
Removes Brainset cards from the docs since they require manual maintenance and do not scale well.


vinamarora8 (commented):
@divyansha1115 amazing. The main README should also be updated as it has links to these cards. Maybe link to the docstring pages over there?

vinamarora8 (approved):
looks good

PR #140, opened 2026-05-16
feat: regularize NLB Maze's behavior signals

vinamarora8:
**Problem:** The behavior timeseries' in NLB maze are _almost regular_ except that they are missing some timesteps. Due to this, we have been using IrregularTimeSeries for them. But IrregularTS is quite bad for behavior signals due to numerical precision errors in sampling.

**Fix:** This PR converts the behavior signals into pure RegularTimeSeries. This is done by inserting `NaN`s in the signal values for the missing timesteps.

### DIFF
--- a/brainsets_pipelines/pei_pandarinath_nlb_2021/pipeline.py
+++ b/brainsets_pipelines/pei_pandarinath_nlb_2021/pipeline.py
@@ -217,6 +217,8 @@ def extract_behavior(nwbfile, trials):
     start_time, end_time = raw_timestamps[0], raw_timestamps[-1]
     num_timesteps = round((end_time - start_time) * samp_rate) + 1
     raw_time_idx = np.round((raw_timestamps - start_time) * samp_rate).astype(int)
+    assert (np.diff(raw_time_idx) > 0).all()
+    # ^ this confirms there are no repeated timesteps
     assert np.isclose(raw_time_idx / samp_rate, raw_timestamps - start_time).all()
     # ^ this confirms that the raw_timestamps are indeed regular relative to start_time
 
--- a/brainsets_pipelines/pei_pandarinath_nlb_2021/pipeline.py
+++ b/brainsets_pipelines/pei_pandarinath_nlb_2021/pipeline.py
@@ -6,9 +6,10 @@
 from argparse import ArgumentParser
 import datetime
 
+import numpy as np
 import h5py
 from pynwb import NWBHDF5IO
-from temporaldata import Data, IrregularTimeSeries, Interval
+from temporaldata import Data, IrregularTimeSeries, Interval, RegularTimeSeries
 import pandas as pd
 
 from brainsets.descriptions import (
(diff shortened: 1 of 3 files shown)


brainsets_pipelines/pei_pandarinath_nlb_2021/pipeline.py
```diff
@@ -203,22 +205,43 @@ def extract_behavior(nwbfile, trials):
...
-        timestamps=timestamps,
+    raw_timestamps = nwbfile.processing["behavior"]["hand_vel"].timestamps[:]
+    raw_hand_pos = nwbfile.processing["behavior"]["hand_pos"].data[:]
+    raw_hand_vel = nwbfile.processing["behavior"]["hand_vel"].data[:]
+    raw_eye_pos = nwbfile.processing["behavior"]["eye_pos"].data[:]
+
+    # These samples are mostly uniformly sampled at 1000Hz,
+    # but have some missing points. Here we insert NaNs at
+    # the missing timesteps to regularize the timeseries
+    samp_rate = 1e3
```
divyansha1115, line 215: maybe we can calculate the sampling rate instead of hardcoding it?

brainsets_pipelines/pei_pandarinath_nlb_2021/pipeline.py
```diff
@@ -203,22 +205,43 @@ def extract_behavior(nwbfile, trials):
...
+    start_time, end_time = raw_timestamps[0], raw_timestamps[-1]
+    num_timesteps = round((end_time - start_time) * samp_rate) + 1
+    raw_time_idx = np.round((raw_timestamps - start_time) * samp_rate).astype(int)
```
divyansha1115, line 218: we should maybe check for duplicate indices in raw_timestamps?

divyansha1115 (commented):
Thank you for fixing this issue. This is a great solution. I think we might want to make this a function since it would likely be useful across all pipelines?

brainsets_pipelines/pei_pandarinath_nlb_2021/pipeline.py
vinamarora8: done!

brainsets_pipelines/pei_pandarinath_nlb_2021/pipeline.py
vinamarora8: Since it's only hardcoded for this dataset, human verified sampling rate should be okay I think?

divyansha1115: sure

brainsets_pipelines/pei_pandarinath_nlb_2021/pipeline.py
```diff
@@ -24,6 +25,7 @@
 )
 from brainsets.taxonomy import RecordingTech, Task
 from brainsets import serialize_fn_map
+from brainsets.utils.misc_utils import calculate_sampling_rate
```
divyansha1115: remove unused import

vinamarora8: done

divyansha1115 (approved):
lgtm

PR #141, opened 2026-05-16
feat: Add utility function `fill_gappy_timeseries`

vinamarora8:
This PR stems from #140.

**Problem:** Many times behavior signals are inherently regular timeseries signals, but have missing values. We have been generally handling them as irregular timeseries. But that is ot great due to numerical precision errors when slicing (sometimes your slice gets 99 points instead of 100). 

**FIx:** Converting to regular timeseries would be great. And one way to do so would be to insert `NaN`s wherever the timestaps are missing. This PR provides a utility to do that.

### DIFF
--- a/brainsets/utils/misc_utils.py
+++ b/brainsets/utils/misc_utils.py
@@ -60,7 +60,7 @@ def fill_missing_timesteps(
     sampling_rate: float,
     gap_value: float = np.nan,
     rtol: float = 1e-3,
-) -> tuple[np.ndarray, np.ndarray | list[np.ndarray]]:
+) -> np.ndarray | list[np.ndarray]:
     """Fill gaps in a regular-but-gappy time series.
 
     Maps an almost-regularly-sampled signal onto a regular grid at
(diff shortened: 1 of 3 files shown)


brainsets/utils/misc_utils.py
```diff
@@ -51,3 +53,103 @@ def calculate_sampling_rate(timestamps: np.ndarray, rtol: float = 1e-3) -> float
...
     return 1.0 / dt
+
+
+def fill_gappy_timeseries(
+    timestamps: np.ndarray,
+    values: np.ndarray | Iterable[np.ndarray],
+    sampling_rate: float | None = None,
+    gap_value: float = np.nan,
+    rtol: float = 1e-3,
+) -> tuple[np.ndarray, np.ndarray | list[np.ndarray]]:
```
vinamarora8: Add a `return_sampling_rate: bool = True`

vinamarora8:
```suggestion
def fill_gappy_timeseries(
    timestamps: np.ndarray,
    values: np.ndarray | Iterable[np.ndarray],
    sampling_rate: float | None = None,
    gap_value: float = np.nan,
    return_sampling_rate: bool = True,
    rtol: float = 1e-3,
) -> tuple[np.ndarray, np.ndarray | list[np.ndarray], float] | tuple[np.ndarray, np.ndarray | list[np.ndarray]]:
```

vinamarora8: Maybe it is unnecessary to return the timestamps, and more necessary to return the sampling_rate

divyansha1115: since we are making passing the sampling rate compulsory I think returning timestamps is okay. For any checks the user might want to perform outside this function timestamps might be helpful

vinamarora8: Technically, if they want timestamps as output, they could pass timestamps into the "values" input of this fn

